PR #5, opened 2022-03-02
Feature/upgrade to floris v3

Bartdoekemeijer:
- Bump FLASC version number to 1.0.
- Upgrade FLASC compatibility to FLORIS v3.0.
- Bug fix by restricting pandas version in requirements.txt and setup.py


bayc:
@Bartdoekemeijer Do we also need to constraint the Pandas version in FLORIS?

Bartdoekemeijer:
@bayc nope, I think `pip` figures that all out. As long as FLORIS is not set on using the latest version of Pandas, it'll pick the most recent version that is compatible both with FLORIS and FLASC (`1.4.0`).

In the unit tests, `FLASC` will automatically `pip install floris==3.0` and pick a compatible `pandas` version, being `1.4.0`.

PR #25, opened 2022-10-04
Use isocalendar() to avoid error in year/week

paulf81:
Is ready to merge

Working on a project recently I had a strange issue where the year/week assignments plot_highlight_data_by_conds would get scrambled.  Posting the question to stack overflow:

https://stackoverflow.com/questions/73911825/how-does-datetime-understand-53rd-week/73912659#73912659

The resolution seems to be a deprecation of .week in favor of .isocalendar().week, so made that change here (don't see any other calls to .week in flasc currently)

This resolves the scrambling issue in my test, and you can see also the improvement in the little demo shown on stack link above.

Finally merging to core.base because I got somehow lapped and have those changes in my local develop.  Hopefully getting these things all merged up will untie some knots...

### DIFF
--- a/flasc/dataframe_operations/dataframe_filtering.py
+++ b/flasc/dataframe_operations/dataframe_filtering.py
@@ -92,7 +92,7 @@ def plot_highlight_data_by_conds(df, conds, ti):
         conds = [conds]
 
     # Convert time arrays to a string with 'year+week'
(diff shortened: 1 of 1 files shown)


paulf81:
Feel good merging this small one without too much discussion, gonna just do it...

PR #26, opened 2022-10-06
Enforce user to specify ti

Bartdoekemeijer:
Improvement / clarifying usage of `floris_tools` functions. FLORIS solutions are typically calculated inside FLASC by first calculating a large set of solutions over a grid (wind direction, wind speed and sometimes turbulence intensity). Then, the FLORIS solutions during FLASC operations are found quickly by interpolating over this precalculated grid. However, this grid does not always cover all relevant conditions, and often it is unclear which turbulence intensity is really assumed when interpolating from the precalculated solutions. This change forces users to manually specify the turbulence intensity assumed by FLORIS for the SCADA data. 

N/A

`floris_tools`

N/A

N/A


Bartdoekemeijer:
Hi @misi9170 and @paulf81. This PR is a small one but wanted to get attention on it again. It forces the user to specify the turbulence intensity in `df` rather than assuming its the same as the value in floris. This just avoids ambiguity and ensures the user knows what assumption is made before proceeding with the analysis. Also, this PR removes `calc_floris` since it's really outdated, as far as I can see. I think we may at some point want to reimplement `calc_floris` with the `ParallelComputingInterface` class and usage of the `timeseries` function in `FlorisInterface`, but I haven't found a need for it currently.

Bartdoekemeijer:
Also I've just merged in the latest develop branch and resolved any conflicts. Good to merge on my side.

flasc/floris_tools.py
```diff
@@ -11,302 +11,45 @@
...
-            df_out = df_out.drop(columns='index')
-
-    t = timerpc() - start_time
-    print('Finished calculating the FLORIS solutions for the dataframe.')
-    print('Total wall time: %.3f s.' % t)
-    print('Mean wall time / function evaluation: %.3f s.' % (t/df.shape[0]))
-    return df_out
-
 
 def interpolate_floris_from_df_approx(
```
misi9170, line 23: Could you add a docstring here to for `interpolate_floris_from_df_approx()`? It doesn't need to be super detailed, but a quick explanation of the main functionality would be good.

flasc/floris_tools.py
```diff
@@ -11,302 +11,45 @@
...
-                np.mean(fi.floris.flow_field.u, axis=(3, 4))
-            )
-            df_out.loc[idx, ti_cols] = np.squeeze(
-                fi.floris.flow_field.turbulence_intensity_field
-            )
-
-    return df_out
-
-
-def calc_floris(df, fi, num_workers, job_worker_ratio=5, include_unc=False,
```
misi9170, line 187: `calc_floris()` gets called by `calculate_wfpower_for_samples()` (from floris_sensitivity_analysis.py), but `calculate_wfpower_for_samples()` is only used in _legacy examples. Should we be considering `calculate_wfpower_for_samples()` redundant, or do we want to continue to support its use by leaving the `calc_floris()` (and, as a result, `_run_fi_serial()`) function available?

misi9170 (changes requested):
Thanks for nudging us on this Bart. I'm fine with requiring the user to specify ti. I've added a couple of requests; see below.

paulf81 (approved):
Hi @Bartdoekemeijer , I approve this in advance of how you and @misi9170 resolve his remaining comments.  I definitely agree with the approaches, both of having a bias toward explicit setting versus implicit where things can get bite us, and then also, I like the approach of intentionally cleaning out functions that we're not using, thank you!!

flasc/floris_tools.py
Bartdoekemeijer: Added a docstring! Let me know if it suffices!

flasc/floris_tools.py
Bartdoekemeijer: I wrote those functions a long time ago and they're really outdated. I don't think they work anymore. I just wanted to keep them around in case we wanted to use them, to have a starting point. I'm also OK with removing them as a whole. We can always go back in the GitHub history to find them anyway.

misi9170: Sounds good. I'm also fine with removing them.

flasc/floris_tools.py
misi9170: That's a very impressive docstring :) Thanks!

misi9170 (approved):

misi9170:
@Bartdoekemeijer I've now approved your changes. Just checking you're happy before I go ahead and merge.

Bartdoekemeijer:
Great, merging!

paulf81:
Thanks @Bartdoekemeijer !

Issue #29, opened 2022-10-18
Remove confidence interval label from energy ratio plot


paulf81:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Currently, energy ratio plots are overly cluttered by labeling of confidence interval bands, opening this issue to prevent labeling of them

### Environment

```markdown
- OS:
- pip environment (can be retrieved with `pip list`):
```

paulf81:
Issue resolved by pull request #30

PR #30, opened 2022-10-19
Make labeling UQ optional in energy ratio plot

paulf81:
Is ready to be merged

Make labeling UQ optional in energy ratio plot, with default set to False 

Issue #29

### DIFF
--- a/flasc/energy_ratio/energy_ratio_visualization.py
+++ b/flasc/energy_ratio/energy_ratio_visualization.py
@@ -20,7 +20,7 @@ import matplotlib.pyplot as plt
 from floris import tools as wfct
 
 
-def plot(energy_ratios, labels=None, label_uq=False):
+def plot(energy_ratios, labels=None, hide_uq_labels=True):
     """This function plots energy ratios against the reference wind
     direction. The plot may or may not include uncertainty bounds,
     depending on the information contained in the provided energy ratio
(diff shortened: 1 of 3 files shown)


flasc/energy_ratio/energy_ratio_visualization.py
```diff
@@ -20,7 +20,7 @@
 from floris import tools as wfct
 
 
-def plot(energy_ratios, labels=None):
+def plot(energy_ratios, labels=None, label_uq=False):
```
Bartdoekemeijer, line 23: I think the variable `label_uq` can be confusing. Users may try to insert a list with the labels for the uncertainty bounds. Perhaps we can change it to `hide_uq_labels=True`?

flasc/energy_ratio/energy_ratio_visualization.py
```diff
@@ -20,7 +20,7 @@
...
     """This function plots energy ratios against the reference wind
     direction. The plot may or may not include uncertainty bounds,
     depending on the information contained in the provided energy ratio
```
Bartdoekemeijer, line 26: Could you update the docstring to include this new option?

Bartdoekemeijer (changes requested):
Thanks for taking care of this! See two small comments.

paulf81:
Good suggestions @Bartdoekemeijer  I took both of them, thank you!

paulf81:
@Bartdoekemeijer made one additional change of including the boolean option in the high-level calling functions and passing them through, using same name and doc string, let me know if this looks ready to go

flasc/energy_ratio/energy_ratio_visualization.py
paulf81: Done!

misi9170 (approved):
Looks fine to me! Another option would be to plot the bounds and mean line in the same legend entry (block showing the bounds with the mean line passing through it), but that is more complicated and likely unnecessary. I'm happy with the option you've provided.

PR #35, opened 2022-10-24
Update df_downsample to improve pandas usage

paulf81:
This is not ready to merged

I have been getting fragmentation warnings and memory issues coming from df_downsample and so wanted to attempt a refactor as we'd done previously for rolling averages.  I think this version removes mutations and the issues leading to fragmentation.  I also wanted to start sketching in the new requirements that time be a column (not the index) at the input/output of the function.  Write now I didn't edit the function itself to make it easy to compare the proposed function in the notebook, both in similarity and timing to what is currently in


Bartdoekemeijer:
Thanks for doing this, Paul! Took a quick peek and looks great so far!

misi9170 (commented):
Hi @paulf81, I've run through your mockup and largely think it looks good. I have pushed a new version up that makes one change: I've replaced the calculation of the linear standard deviation with a circular one, which is consistent with the changes made to `df_movingaverage()`. 

This slows down your implementation fractionally, but is still considerably faster than the original implementation, and
- seems to decrease mismatch between the current implementation and proposed one
- is, in my opinion, a better way of computing the standard deviation

paulf81:
Nice @misi9170 , I agree with your approach and thank you for checking timing and showing the lower error, I feel good about moving forward with this.  I was going to try to wrap up the block bootstrapping, would you mind wrapping up this one?

misi9170:
@paulf81 yes, I'll make the proposed changes to time_operations.py.

misi9170:
@paulf81 This is now ready for confirmation to merge.

Issue #39, opened 2022-11-07
Block bootstrapping


paulf81:
Add block bootstrapping option to FLASC

paulf81:
Resolved by pull request #40

PR #40, opened 2022-11-07
Feature/block bootstrap

paulf81:
is NOT ready to be merged

This is a first draft implementation of block bootstrapping.  In our current bootstrapping, we sample all R rows from a dataframe with R totals rows (with replacement) N times.  In the proposed block bootstrapping method, the data are divided into D consecutive blocks (with the 0th block including any residual rows at the end), and we sample D of these blocks (with replacement) N times instead.   I think theory is roughly like:

https://en.wikipedia.org/wiki/Bootstrapping_(statistics)#Block_bootstrap

But essentially since we have 1/10 minute samples, these might relate a lot in time, so removing whole blocks from iterations could help to better show underlying uncertainty

I added a demonstration to the example: energy_ratio_vs_wd_for_single_df, but should work also with suite

Issue #39

### DIFF
--- a/flasc/energy_ratio/energy_ratio.py
+++ b/flasc/energy_ratio/energy_ratio.py
@@ -300,8 +300,7 @@ class energy_ratio:
         N=1,
         percentiles=[5.0, 95.0],
         return_detailed_output=False,
-        block_bootstrapping=False, 
-        num_blocks=10
+        num_blocks=-1
     ):
         """This is the main function used to calculate the energy ratios
         for dataframe provided to the class during initialization. One
@@ -364,6 +363,10 @@ class energy_ratio:
                 direction and wind speed bin, among others. This is
                 particularly helpful in figuring out if the bins are well
                 balanced. Defaults to False.
+            num_blocks (int, optional): Number of blocks to use in block
+                boostrapping.  If = -1 then don't use block bootstrapping
+                and follow normal approach of sampling num_samples randomly
+                with replacement.  Defaults to -1.
 
         Returns:
             energy_ratios ([pd.DataFrame]): Dataframe containing the found
@@ -408,7 +411,6 @@ class energy_ratio:
             N=N,
             percentiles=percentiles,
             return_detailed_output=return_detailed_output,
-            block_bootstrapping=block_bootstrapping, 
             num_blocks=num_blocks
         )
         if return_detailed_output:
@@ -509,8 +511,7 @@ def _get_energy_ratios_all_wd_bins_bootstrapping(
     N=1,
     percentiles=[5.0, 95.0],
     return_detailed_output=False,
-    block_bootstrapping=False, 
-    num_blocks=10
+    num_blocks=-1
 ):
     """Wrapper function that calculates the energy ratio for every wind
     direction bin in the provided dataframe. This function wraps around
@@ -552,6 +553,10 @@ def _get_energy_ratios_all_wd_bins_bootstrapping(
             direction and wind speed bin, among others. This is
             particularly helpful in figuring out if the bins are well
             balanced. Defaults to False.
+        num_blocks (int, optional): Number of blocks to use in block
+            boostrapping.  If = -1 then don't use block bootstrapping
+            and follow normal approach of sampling num_samples randomly
+            with replacement.  Defaults to -1.
 
     Returns:
         energy_ratios ([pd.DataFrame]): Dataframe containing the found
@@ -600,7 +605,6 @@ def _get_energy_ratios_all_wd_bins_bootstrapping(
                 N=N,
                 percentiles=percentiles,
                 return_detailed_output=return_detailed_output,
-                block_bootstrapping = block_bootstrapping, 
                 num_blocks = num_blocks
         )
         if return_detailed_output:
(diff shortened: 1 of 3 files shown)


paulf81:
@misi9170 same question to you

flasc/energy_ratio/energy_ratio.py
```diff
@@ -299,6 +300,8 @@ def get_energy_ratio(
         N=1,
         percentiles=[5.0, 95.0],
         return_detailed_output=False,
+        block_bootstrapping = False,
```
Bartdoekemeijer, line 303: Remove spaces around equal sign for consistency

flasc/energy_ratio/energy_ratio.py
```diff
@@ -405,6 +408,8 @@ def get_energy_ratio(
             N=N,
             percentiles=percentiles,
             return_detailed_output=return_detailed_output,
+            block_bootstrapping = block_bootstrapping,
```
Bartdoekemeijer, line 411: Same here

flasc/energy_ratio/energy_ratio.py
```diff
@@ -504,6 +509,8 @@ def _get_energy_ratios_all_wd_bins_bootstrapping(
     N=1,
     percentiles=[5.0, 95.0],
     return_detailed_output=False,
+    block_bootstrapping = False,
```
Bartdoekemeijer, line 512: and here

flasc/energy_ratio/energy_ratio.py
```diff
@@ -504,6 +509,8 @@ def _get_energy_ratios_all_wd_bins_bootstrapping(
     N=1,
     percentiles=[5.0, 95.0],
     return_detailed_output=False,
+    block_bootstrapping = False, 
+    num_blocks = 10
```
Bartdoekemeijer, line 513: I wonder if somehow we can make this more uniform. Essentially, regular bootstrapping is block bootstrapping with num_blocks==num_samples, correct? Maybe that can simplify things like that and that comparison can also work as a unit test.

Then, we can collapse the two options `block_bootstrapping` and `num_blocks` just to `num_bootstrapping_blocks`, with `-1` being the default option meaning `num_blocks=num_samples`? Or maybe that would all become too ambiguous. What do you think?

flasc/energy_ratio/energy_ratio.py
```diff
@@ -655,11 +666,33 @@ def _get_energy_ratio_single_wd_bin_bootstrapping(
     if N <= 1:
         results_array = np.array([energy_ratio_nominal] * 3, dtype=float)
     else:
+
+        # If using block-bootstrapping, set up blocks
+        if block_bootstrapping:
+            block_list = list(range(num_blocks))  # List of all block names
```
Bartdoekemeijer, line 672: I haven't tried this, but I wonder how things work if you have 10 samples and split it over 3 blocks. It should split it into 3 vectors with length 4, 3 and 3, respectively. And if you have 8 samples and split it into 3 blocks, it should have three vectors with length 3, 3 and 2. Do those work out, currently? My thought with the current line of code ` block_length = int(len(block_indices) / num_blocks)` was that this may not be resolved properly.

Maybe consider `np.split_array` is a good function for this. For example
```
sample_indices = np.arange(0, num_samples, dtype=int)
block_indices = np.array_split(sample_indices, num_blocks)
```

flasc/energy_ratio/energy_ratio_suite.py
```diff
@@ -315,6 +315,8 @@ def get_energy_ratios(
         percentiles=[5.0, 95.0],
         balance_bins_between_dfs=True,
         return_detailed_output=False,
+        block_bootstrapping = False,
```
Bartdoekemeijer, line 318: remove spaces adjacent to equal sign

flasc/energy_ratio/energy_ratio.py
```diff
@@ -655,11 +666,33 @@ def _get_energy_ratio_single_wd_bin_bootstrapping(
...
+
         # Get a bootstrap sample of range
         bootstrap_results = np.zeros(N)
         bootstrap_results[0] = energy_ratio_nominal
         for i in range(1, N):
-            df_randomized = df_binned.sample(frac=1, replace=True).copy()
+            if not block_bootstrapping:
+                df_randomized = df_binned.sample(frac=1, replace=True).copy()
+            else:
+                test_blocks = choices(block_list, k = num_blocks) # Choose a set of blocks
```
Bartdoekemeijer, line 686: I like this and I think this can also work for regular bootstrapping if num_blocks == num_samples.

Bartdoekemeijer (changes requested):
Thanks for doing this, Paul! See my comments.

flasc/energy_ratio/energy_ratio.py
paulf81: Done!

flasc/energy_ratio/energy_ratio.py
paulf81: Done!

flasc/energy_ratio/energy_ratio_suite.py
paulf81: Done!

flasc/energy_ratio/energy_ratio.py
paulf81: I like it in principle, my only doubt is if the block-bootstrapped version of normal bootstrapping with num_blocks=num_samples is slower than the present implementation?  I should test, if that's ok, I think it makes sense.  What do you think @misi9170 ?

flasc/energy_ratio/energy_ratio.py
paulf81: I like that, I'll give it a shot!

flasc/energy_ratio/energy_ratio.py
Bartdoekemeijer: I agree, would be good to make sure things don't slow down!

paulf81: hi @Bartdoekemeijer  I tried this by putting in this code block around line 670 in energy_ratio.py:

```
        if not block_bootstrapping:
            num_blocks = df_binned.shape[0]
            block_bootstrapping = True
```
Unfortunately it was about 10x slower, I think maybe because the other call functions through built-in pandas (numpy) methods it gets a big speed up.   So maybe we keep the logic as is? 

 Still though, I think we could have the function inputs act as you like, instead of the seperate boolean, just haave num_blocks default to -1, meaning no blocks, does that work for you?

flasc/energy_ratio/energy_ratio.py
paulf81: hi @Bartdoekemeijer  I remember what I did here, I start out by assigning everything to the first (0th block) with this line:

```
block_labels = np.zeros(len(block_indices)).astype(int) # Labels to assign each index
```

then I assign the blocks here:

```
for b_i in range(num_blocks):
    block_labels[block_length * b_i:block_length * (b_i+1)] = b_i
```

If the things don't divide evenly the leftovers get assigned to the first (0th block) which I guess could make that block a little bigger than the others but it gives some wrapping around, which I thought could be desirable.  But maybe your method is still preferable since it makes sure things are more even

flasc/energy_ratio/energy_ratio.py
Bartdoekemeijer: Yes, that works! Thanks for checking, and I agree that a 10x slowdown is not worth it.

misi9170: I'd say feel free to merge if this does not affect current code functionality. Given the slowdown, it's best to keep the bootstrapping and block bootstrapping separate.

paulf81:
hi @Bartdoekemeijer and @misi9170 I think I've updated now as we discussed, block-bootstrapping is initiated via setting  num_blocks something other than the default of -1.  
I updated the example and also updated the docstrings.
And I added a valueError check on num_blocks
@misi9170  what do you think, good to merge?

Issue #38, opened 2022-10-31
[Feature]: Reduce precision of dataframes


Bartdoekemeijer:
### Description

FLASC often works with large amounts of data. This becomes particularly apparent if the user is working with 1Hz data rather than 10-minute averages. To reduce file size and memory usage, dataframes can be formatted to an appropriate precision.

Namely:
- Almost all floats can be saved as 32-bit floats rather than 64-bit floats, since the additional precision will drown in measurement noise anyway.
- According to @paulf81 strings are better saved as categories or chars, not just objects
- True/False columns should be saved as booleans, not integers

This can generally reduce the size of the dataframes by typically 50%, sometimes more. A simple function can be written and added to `dataframe_manipulations` to reduce the precision of any dataframe.

PR #41, opened 2022-11-07
Feature/type downsample

paulf81:
Is NOT ready to be merged

Put in a very initial version of the downsample types function and a companion example, a number of caveats though
- not sure where the function should live
- not sure where the example should live
- function only handles floats for the moment

  Code based on stuff we've used previously and I tried to remove any mutations

Issue #38

### DIFF
--- a/examples/demo_downsample.py
+++ b/examples/demo_downsample.py
@@ -2,7 +2,7 @@ import os
 import pandas as pd
 import numpy as np
 
-from flasc.time_operations import downsample_types
+from flasc.dataframe_operations.dataframe_manipulations import df_reduce_precision
 
 
 def load_data():
@@ -31,7 +31,7 @@ if __name__ == "__main__":
     df = load_data()
 
     # Downsample the types
-    df_out = downsample_types(df, verbose=True)
+    df_out = df_reduce_precision(df, verbose=True)
 
     print()
     print('---------------')
--- a/examples/demo_downsample.py
+++ b/examples/demo_downsample.py
@@ -1,0 +1,40 @@
+import os
+import pandas as pd
+import numpy as np
+
+from flasc.time_operations import downsample_types
+
+
+def load_data():
+    # Load dataframe with scada data
+    root_dir = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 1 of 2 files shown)


Bartdoekemeijer:
Hi Paul. Check this out: https://github.com/NREL/flasc/blob/8b6a21eee80f8eb20481ee4b3d4231d7c0c09699/flasc/dataframe_operations/dataframe_manipulations.py#L685

I forgot that I had written a function for this a while ago. It does a couple of additional things:
- Simplify 64-bit float columns to 32-bit float
- Simplify floats to integers if possible
- Simplify higher-bit integers to lower-bit integers (down to 8-bit) if possible
- Simplify from integer column to bool column if it contains only 0s and 1s

Additionally, I'd like to add functionality that recognizes strings, determines whether they can be better sorted as categories (like you previously suggested), and format them accordingly.

Bartdoekemeijer:
> @Bartdoekemeijer huge miss on my part! Sorry for that! I've now removed the duplicate function and updated the new demo to point to your existing code and show it reduces mem usage by about 50%. Then I'll leave the pull open where we can add the other operations you mention above. Maybe also I need to move the location of this example, do you have a preferred folder?

Ha no worries, an even bigger oversight on my part since I wrote that function, haha! Yes, sounds good! Maybe we can just have the reduce operation in one of the raw_data_processing steps, early on. Maybe in `a_00_initial_download.py` or `a_01_to_common_format_df.py`?

paulf81:
@Bartdoekemeijer  this could mean it will be in the new repository then right?  I mean like with Issue #48

Bartdoekemeijer:
> @Bartdoekemeijer this could mean it will be in the new repository then right? I mean like with Issue #48
Yes, sounds good to me! I'm currently building such a template repository but haven't made it public yet. I'd like to get FLASC available through PyPI first. That will allow me to link the template repository to a specific version of FLASC. I'll open a separate issue for that.

Bartdoekemeijer:
I think we can close this now that we have https://github.com/NREL/flasc_cookiecutter_template.

paulf81:
Agreed I will close it now

PR #47, opened 2022-11-16
Add handling for plotting a subset of turbines in waking directions vis

misi9170:
Ready to be merged.

In current version, the only way to handle plotting the layout or waking directions of a subset of the farm (using `plot_layout_with_waking_directions()` is to pass a FlorisInterface with a layout for that subset. This modification allows users to specify a subset of turbine indices which they would like to see waking directions for, utilizing an existing mechanism already available in `plot_layout_only()`. Also allowing user to define the font size for labeling the wind directions/distances.

To specify a subset of turbines of interest, the `layout_plotting_dict` argument provided to `plot_layout_with_waking_directions()` should contain a key `"turbine_indices"` that is a list of the turbine indices to be plotted. Font size for labels is set with new argument `wake_label_size`.

Default behavior unchanged.

Test: no change to the wake directions plot produced by examples/wake_steering_design/00_analyze_single_ws_vs_range.py


paulf81:
hi Misha, I think this is a great idea!  Couple of quick comments:

- The current pr is pointing to main, could you point at develop?
- Could you catch up feature/layout_viz to develop (some small changes in codecov etc)

misi9170:
@paulf81 My mistake on the base branch! And yes, thought I was all up to speed with develop but apparently not---that's now pulled and merged. I've also reverted one unnecessary step adding an input to the inner `put_label` function.

paulf81:
ok that looks good!  my last thought is we could probably use a new example, that just shows how to do each of the plot layout types, I'm happy to propose something?

misi9170:
Sure, sounds good, please do!

paulf81:
Ok I've just pushed up an example of making different types of layout plots, does this work/look ok to you @misi9170 ?

misi9170:
The example looks good. I've just added one more option in the final plot to demonstrate the selection of a subset of turbines to plot waking information for. Ready to merge?

paulf81:
Yes!

Issue #52, opened 2022-11-28
[BUG] return_index_mapping option needed in df_downsample


misi9170:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

A recent feature that updated `df_downsample()` removed the `return_index_mapping` flag and the corresponding `data_indices` output, breaking this functionality for some users.

### Expected Behavior

`df_downsample()` should take `return_index_mapping` as a flag and, when this flag is True, return an array `data_indices` that maps indices in the high resolution input data to indices in the downsampled output data.

### Environment

```markdown
- OS:
- pip environment (can be retrieved with `pip list`):
```

PR #53, opened 2022-11-28
Reinstate return_index_mapping in df_downsample

misi9170:
Recent updates to `df_downsample()` (in time_operations.py) removed functionality for providing information about which indices from the full resolution dataframe went into each index of the downsampled dataframe. This broke some functionality for some users. 

This bugfix addresses this issue by reinstating the `return_index_mapping` flag to return the `data_indices` output in `df_downsample()`. Note that, at this point, `df_movingaverage()` still does _not_ return `data_indices`, although the `return_index_mapping` flag remains an input option. This may be addressed in a future bug fix if needed. 

Bug is reported in issue #52. Change would close this issue.

Changes made to `df_downsample` and test code.

### DIFF
--- a/flasc/time_operations.py
+++ b/flasc/time_operations.py
@@ -140,7 +140,8 @@ def df_downsample(
     window_width=td(seconds=60),
     min_periods=1,
     center=False,
-    calc_median_min_max_std=False
+    calc_median_min_max_std=False,
+    return_index_mapping=False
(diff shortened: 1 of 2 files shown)


paulf81:
ok!  I checked this with my dataset and fixes issues so going ahead and merging, thank you @misi9170 !

PR #54, opened 2022-11-28
catch case of small number of points

paulf81:
Yes

Add check for small number of points in block bootstrapping (less than number of blocks) and in that case revert to using each point as block


misi9170:
This looks good to me. Any chance you want to print a warning if this happens so that the user is notified that something will be different? E.g.        
```
if num_blocks > df_binned.shape[0]:
            num_blocks = df_binned.shape[0]
            print("Warning: num_blocks reduced to match the size of the binned data.")
```

misi9170 (approved):
Fair enough regarding avoiding being too spammy. Feel free to merge.

paulf81:
Done, thanks @misi9170 !

Issue #31, opened 2022-10-20
Plot energy ratio gain


paulf81:

paulf81:
hi @Bartdoekemeijer, @ejsimley and @misi9170   I'm assigning this one to myself but was hoping to ask your input on getting started.  Basically, I've been looking at how we had done this previously (calculating and plotting the gain of one energy ratio to another) and was thinking the right place to work this in would be into the energy_ratio_suite framework since it only makes sense if you have two things.  In the sort of standard case you'd have a df now for scada_baseline, scada_control, floris_baseline, floris_control, and maybe we'd have a function option to plot the gains and you just have indicate which ones to make a gain on, does that make sense?

paulf81:
btw for reference I've been looking at FLORIS v2.3 which had this functionality

misi9170:
as you suggested, energy_ratio_suite looks like the place for this based on its docstring. I'll let Bart confirm though!

Bartdoekemeijer:
Yes agreed! And the indicator to decide which to make a gain on should be a list. Namely, you should specify for each case to which other case should be normalized. That'll allow you to plot scada_control normalized by scada_baseline and floris_control normalized by floris_baseline in a single function call.

misi9170:
Closing as #58 was merged.

PR #58, opened 2022-12-14
Feature/energy ratio gain

paulf81:
It works, but probably there's some cleaning up to do first.  Also, I'm not 100% I did everything accurately, it gets a little confusing to think through.  I think it's right, but would greatly appreciate a review of the code for accuraccy

Adds the ability to compute the gain in energy ratio, including bootstrapping (normal or block).  For now, the assumption in the suite code is that if you provide 2, 4 etc., dataframes, then the order goes denom_df_1, num_df_1, denom_df_2, num_df_2., etc., not sure if this is the best way here.

Couple of additional thoughts:
- I was thinking a potentially simpler approach would be to compute the energy ratio and uncertainties of each df seperately and somehow combine the individual uncertainties into the uncertainty of the ratio, but I'm not sure how this works.  Also, with this more thorough approach, one nice advantage is that weights can be set by the minimum of the pair, emphasizing bins with the most points in the less fill dataframe

- The example, wake_steering_example.py reminds me that it'd be nice, someday, to prove something.  Basically, adding noise changes the distribution of the gain, and noticably reduces the peak gain, however, on some level it shouldn't effect the overall gain in energy production in total across ws/wd I would think?  

Issue #31 

energy_ratio_suite.py

### DIFF
--- a/examples/energy_ratio/wake_steering_example.py
+++ b/examples/energy_ratio/wake_steering_example.py
@@ -42,7 +42,7 @@ if __name__ == "__main__":
     # Create a time history of points where the wind speed and wind direction step different combinations
     ws_points = np.arange(5.0,10.0,1.0)
     wd_points = np.arange(250.0, 290.0, 1.,)
-    num_points_per_combination = 5 # How many "seconds" per combination
+    num_points_per_combination = 5 # 5 # How many "seconds" per combination
 
     # I know this is dumb but will come back, can't quite work out the numpy version
     ws_array = []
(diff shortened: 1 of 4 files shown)


paulf81:
Making this recent example with the gain got me thinking something interesting that in this toy problem, we know definitionally that the total energy gain in the non-noisy and the noisy case are in total identical, but the noisy case looks much less:

image


And I think this happens a bit in practice with field data.  It made me think if energy ratio of energy ratios is the wrong way to visualize the improvement of wake steering / wfc, and maybe the difference in energy ratios is more clear, easier to total up?

image

For example, in this example we know the energy production of the downstream turbine, in both cases rises from 967879447.3303714 to 1030753035.8542225 (up 6.5%), I wonder if we can show this value can be recovered from the normal and noisy energy ratio gains if we use difference rather than ratio?  @misi9170 @ejsimley @Bartdoekemeijer ?

Bartdoekemeijer:
Thanks for sharing this analysis, Paul! My feeling is that a difference in energy ratios should be less sensitive to noise than a ratio of energy ratios. Especially thinking about small energy ratio values, a small absolute difference can trigger a much larger relative difference.

I'm not sure about reverse-engineering the actual uplift of wake steering by somehow removing the noise. The step from non-noisy to noisy seems like some sort of Gaussian smoothing. If we can figure out the Gaussian smoothing filter (standard deviation) we need to go from baseline to noisy in this toy example, perhaps we can reverse this operation?

paulf81:
I was thinking about this a little over the break and I still think the difference versus ratio is interesting for visualization, maybe a procedure for estimating overall change could be like:
 - [1] Compute ranges of, "ratio of energy ratios" across all wd bins
 - [2] Compute the energy production for the test turbine in the baseline condition for all wind directions
 - [3] Now the expected energy production of the test turbine in the control condition is the dot product of the above
 - The ratio of the sum of [2] and the value in [3] is expected total uplift,

ejsimley:
Hi @paulf81 I think the approach you described a few days ago is generally correct. One concern though is that I believe the energy ratios for baseline and wake steering modes are balanced to represent the combined wind speed distribution during baseline and wake steering control (in other words, the energy in each wind speed bin is normalized so it contributes the same relative amount to the total energy ratio in the baseline and wake steering cases even if baseline and wake steering periods had different wind speed distributions). At least I think that is the case in FLASC? In that case instead of multiplying the ratio of energy ratios by the baseline energy production of the test turbine, it might be better to weight the test turbine's baseline energy in each wind speed bin using the same weights as in the energy ratio before multiplying.

Another approach would be to take the dot product between the difference in the energy ratios (wake steering - baseline) and the total energy of the reference turbine. This way, I think the reference turbine's energy would already match the wind speed distribution assumed in the energy ratios. 

I also have some ideas about computing the total energy gain using energy ratios binned by wind direction and wind speed, so that way it is easier to use either the wd/ws bin weights during the experiment or weights based on a wind rose to estimate the long-term energy gain.  

Not sure if this made sense, so let's discuss sometime.

paulf81:
Thanks Eric.  It all does get a little confusing, but I'm pretty sure I understand what you're saying and agree with you.  I think then it will be hard to do this right, without embedding this within the class (which we'd do eventually) so we have access to the right weights right?

But I also like where you're going with maybe the way to do total is to set aside the by wd-only gains and just go ws/wd bin by bin in a separate operation.  Is that what you're saying?  I can see that making a lot of sense and would be great to talk to you more about it

misi9170:
Resolves #31

PR #66, opened 2023-02-11
Add table based analysis

paulf81:
NOT Ready to be merged

@misi9170 and @ejsimley  I have been discussing how best to compute total uplift with uncertaintity bounds.  We've talked about reshaping things into an alternative approach where every value is placed in a 2d (wd/ws) (I've gone 3d wd/ws/turbine following FLORIS ordering) table, and trying to approach confidence bounds using some sort of gaussian.  The present pull request starts this but doesn't finish because there are still some questions like:

- How to assign bounds to bins with very low number of points?  Assume the maximum?  Assume some default?
- If we model the bins as gaussians, what is the gaussian bounds of their sum?  Of their ratio?
- I assume frequency comes from the mininum, to be sure we always 0-weight bins that dont have any points in one of the cases, this could also be done in other ways

I also haven't fully checked all the math so we'll want to develop some tests to make sure everything is computing

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yaml
+++ b/.github/workflows/continuous-integration-workflow.yaml
@@ -43,5 +43,6 @@ jobs:
       uses: codecov/codecov-action@v3
       with:
         token: ${{ secrets.CODECOV_TOKEN }}
+        files: ./coverage.xml
         fail_ci_if_error: true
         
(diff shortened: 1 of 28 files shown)


misi9170:
@paulf81 I'm going to close this as no longer relevant since #110 went in. Please reopen if you would like to.

Issue #64, opened 2023-02-10
[Feature]:  Increase usage of OpenOA


Bartdoekemeijer:
### Description

OpenOA seems like a very mature tool for basic SCADA analysis and operations. FLASC currently has a set of tools that do very similar things as OpenOA and has minor dependencies on OpenOA, but could be much better integrated. Putting in this as a placeholder/reminder.

### Related URLs

https://github.com/NREL/OpenOA

paulf81:
I'm sure @ejsimley agrees!  Something we've started doing in FLORIS we might try here is opening a discussion thread where we can have a discussion (that is saved) on which features it could make sense to migrate

paulf81:
Started discussion #65

PR #68, opened 2023-02-15
Fix NaN handling in calc_wd_mean_radial

misi9170:
### Issue
`calc_wd_mean_radial()` produces some inconsistent behavior. When the `angles_array_deg` input is a numpy array that contains NaNs, the output mean wind direction is NaN. However, when `angles_array_deg` is a dataframe, with only a few NaN values, the output ignores the NaNs, and when `angles_array_deg` contains entire rows/columns of NaNs, the output is 0. 

### Solution
I propose the following approach:
- If some, but not all, wind directions in the mean are NaNs, those values are ignored and the mean is over the non-NaN values
- If all wind directions in the mean are NaNs, NaN is returned.

Note that this is slightly different than current behavior, which:
- returns NaN when any NaNs are present if `angles_array_deg` is a numpy array; and returns the mean over non-NaN values if `angles_array_deg` is a dataframe.
- returns NaN when all wind directions are NaN if `angles_array_deg` is a numpy array; and returns 0 when all wind directions are NaN if `angles_array_deg` is a dataframe.

### Steps to recreate issue

Run the following code on the current develop branch and compare to this bugfix branch:
```
from flasc import circular_statistics as cs
import numpy as np
import pandas as pd

print(np.nansum(np.array([3, 2, np.nan]))) # Intuitive
print(np.nansum(np.array([np.nan, np.nan, np.nan]))) # Not particularly intuitive

# generate data centered around 0
np.random.seed(1)
X = 360 + np.random.randn(5, 3)*5

## Behavior for numpy arrays

print(X)
# Normal behavior
print(cs.calc_wd_mean_radial(X))
print(cs.calc_wd_mean_radial(X, 1))

# Behavior when NaNs present
X_nans = X.copy()
X_nans[:,1] = np.nan
X_nans[1,:] = np.nan

print(X_nans)
# All nans due to presence of at least one nan
print(cs.calc_wd_mean_radial(X_nans)) 
print(cs.calc_wd_mean_radial(X_nans, 1)) 

## Behavior when data is a pandas dataframe
df = pd.DataFrame(X)

print(df)
# Runs as expected when NaNs not present
print(cs.calc_wd_mean_radial(df)) 
print(cs.calc_wd_mean_radial(df, 1)) 

df_nans = pd.DataFrame(X_nans)
print(df_nans)
# Runs as expected when not all values are NaN, but produces zeros when all 
# values are NaN
print(cs.calc_wd_mean_radial(df_nans)) 
print(cs.calc_wd_mean_radial(df_nans, 1)) 
```

I could also turn the above (or something similar) into a unit test for `calc_wd_mean_radial()`.

### DIFF
--- a/flasc/circular_statistics.py
+++ b/flasc/circular_statistics.py
@@ -18,17 +18,19 @@ from floris.utilities import wrap_360
 
 def calc_wd_mean_radial(angles_array_deg, axis=0):
     
-    # Prevent calculation if all inputs are NaNs
-    if np.isnan(angles_array_deg).all():
-        return np.nan
+    # Generate mask for all NaN rows/columns
+    nan_mask = np.isnan(angles_array_deg).all(axis=axis)
     
     # Use unit vectors to calculate the mean
     wd_x = np.cos(angles_array_deg * np.pi / 180.)
     wd_y = np.sin(angles_array_deg * np.pi / 180.)
 
-    mean_wd = np.arctan2(np.sum(wd_y, axis=axis),
-                         np.sum(wd_x, axis=axis))
+    mean_wd = np.arctan2(np.nansum(wd_y, axis=axis),
+                         np.nansum(wd_x, axis=axis))
     mean_wd = wrap_360(mean_wd * 180. / np.pi)
+    
+    # Replace values based on all NaNs with NaN
+    mean_wd[nan_mask] = np.nan
 
     return mean_wd
 
--- a/flasc/circular_statistics.py
+++ b/flasc/circular_statistics.py
@@ -17,6 +17,11 @@ from floris.utilities import wrap_360
 
 
 def calc_wd_mean_radial(angles_array_deg, axis=0):
+    
+    # Prevent calculation if all inputs are NaNs
+    if np.isnan(angles_array_deg).all():
+        return np.nan
+    
     # Use unit vectors to calculate the mean
     wd_x = np.cos(angles_array_deg * np.pi / 180.)
     wd_y = np.sin(angles_array_deg * np.pi / 180.)


Bartdoekemeijer:
Nice find @misi9170! Perhaps we can add an option for NaN handling, similar to `scipy.stats.circmean`, defaulting to `omit` as you suggest? See https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.circmean.html.

misi9170:
@paulf81 @Bartdoekemeijer OK to merge?

paulf81:
I was just about to say, but won't this be slower?  But happy to hear it's actually faster!  Good to go from my end

paulf81 (approved):

misi9170:
@paulf81 Thanks for calling that out---I was wrong. I went back and checked, and I had been comparing the times incorrectly. In fact, this proposed implementation (using scipy's `circmean`) is about 15% _slower_ than the the current implementation. In my opinion, that's a reasonable price to pay for simplicity, but I'm also happy to revert to the explicit calculation. 

For reference, this operation takes about 3.5 seconds to calculate 1,000,000 means, each containing 100 points, which I think is still fast enough to not be a major hold up for dealing with SCADA records.

Bartdoekemeijer:
15% slower for the sake of simplicity seems fine to me!

Bartdoekemeijer (approved):

misi9170:
Great---squashing and merging.

paulf81:
Thanks guys, and I think agree, simplicity is a good goal.  Also was thinking the more we have a practice of using tools versus reinventing probably then we'll also set ourselves up to benefit from improvements made to those tools, so probably in addition to simplicity, this is a good decision from the point of view of good practice, like, don't write any functions that exist already in numpy/scipy.  I know this is already merged was just thinking of adding this point to maybe nudge future decisions in this same way.  Thanks @misi9170  and @Bartdoekemeijer !

Bartdoekemeijer:
Totally agree @paulf81! I think this ties in with #64 too. The more we can rely on OpenOA, the leaner this code becomes!

Issue #57, opened 2022-12-13
[Feature]: Indicate turbine dependencies for case to case testing


misi9170:
### Description

For large farms, we need a way of determining which (upstream) turbines a given turbine's power depends on.

Should fit naturally in floris_tools.py

Bartdoekemeijer:
You could look at

https://github.com/NREL/floris/blob/39c466000b1874e06a6f58da9c30bb877fc8d4d2/floris/tools/optimization/yaw_optimization/yaw_optimization_tools.py#L21

and

https://github.com/NREL/floris/blob/6e856e54753f290de83eeeb74b7bc97c35fe22ad/floris/tools/optimization/yaw_optimization/yaw_optimization_tools.py#L20

These functions rely on a simple geometrical wake model to quickly determine which turbines fall inside the wake of another turbine.

misi9170:
@Bartdoekemeijer thanks for pointing these out! I'll take a look before implementing anything. One 'feature' for this that Paul and I have been discussing is that we'd like to set a threshold for dependency (user defined, but default at 1% of rated power), so turbines far downstream (but still technically in the wake) of an upstream turbine are not considered dependent.

paulf81:
Thanks @Bartdoekemeijer !  These are a good pointer, but do agree with @misi9170 this new method has some novelties.  Also it's a bit inverted of these, it's making a list of upstream turbines

Bartdoekemeijer:
I see! The methods above are really simplified and do not include any kind of quantitative wake loss. You could include a downstream distance threshold here for dependency, e.g., 30 rotor diameters. I can imagine having a threshold for wake loss dependency would require more complicated methods and could therefore be a lot slower but more accurate. Depending on your use case, I can see either work!

paulf81:
I think that is a nice way to think of it, both are options, and I think the geometric approach will be much faster, but the wake model approach will allow criteria based on relative level of impact

misi9170:
Closed by PR #70

PR #70, opened 2023-02-23
Code for determining which turbines impact the power of a given turbine

misi9170:
Ready for review and merge.

New functionality added to floris_tools.py to compute which turbines impact and depend on a specified turbine, based on their variations in power predicted by FLORIS.

Tasks:
- [x] Implement code to determine which turbines _impact_ a specified turbine
- [x] Implement code to determine which turbines _depend on_ a specified turbine
- [x] Implement wrappers for both of the above to run over all turbines [Note: _dependencies_ are available per turbine. _impacting turbines_ are available only as a full set, due to code structure]
- [x] Create example that shows usage
- [x] Create tests

### DIFF
--- a/examples/layout/turbine_dependencies.py
+++ b/examples/layout/turbine_dependencies.py
@@ -24,6 +24,14 @@ print("Turbines that depend on T002 at 226 degrees:",
       depend_on_2[round(226/2)]
      )
 
+# Can also return all influences as a matrix for other use (not ordered)
+depend_on_2, influence_magnitudes = fsatools.get_dependent_turbines_by_wd(
+    fi, 2, check_directions, return_influence_magnitudes=True)
+print("\nArray of all influences of T002 has shape (num_wds x num_turbs): ", 
+      influence_magnitudes.shape)
+print("Influence of T002 on T006 at 226 degrees: {0:.4f}".format( 
+      influence_magnitudes[round(226/2), 6]))
+
 df_dependencies = fsatools.get_all_dependent_turbines(fi, check_directions)
 print("\nAll turbine dependencies using default threshold "+\
       "(first 5 wind directions printed):")
--- a/examples/layout/turbine_dependencies.py
+++ b/examples/layout/turbine_dependencies.py
@@ -6,6 +6,15 @@ from flasc import floris_tools as fsatools
 from flasc import visualization as fsaviz
 
 from floris import tools as wfct
+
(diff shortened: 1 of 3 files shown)


misi9170:
Addresses #57

Bartdoekemeijer:
These functions are super helpful, Misha. Have you already figured out how you want to do (1)?

Bartdoekemeijer:
Maybe to continue this conversation, I'd like the same functionality and I had a look in the FLORIS code, here:
https://github.com/NREL/floris/blob/1d84538c334a502c6ad7df48b8cc2309d6a6436d/floris/simulation/solver.py#L161

which introduces the variable `velocity_deficit`. The function is called in a loop from 0 to n_turbines. The variable has dimensions `(n_wind_directions, n_wind_speeds, n_wind_turbines, n_rotor_points, n_rotor_points)`. So you get one of these variables for each turbine, and it shows the wake losses it causes at all downstream and upstream turbines. With some basic logic, you can derive which turbines are impacting which downstream turbines, and hence vice versa, without rerunning any FLORIS code.

Note that this solution does not take care of wake summation so there may be some small differences between this and your method, but perhaps still relatively small. Could be worth considering, in case you hadn't made up your mind yet!

misi9170:
@Bartdoekemeijer thanks for the comment here! I have now added a first pass for the first task, which simply rearranges the output of the "dependent turbine" function. It's a bit clunky though, because the returned impacting turbines are not ordered. Still, a possible fix is to generate this as a class and assign the size of the impacts (found in the `dependent` function to memory so that they can be later used for ordering in the `impacting` function). I'll also think about your suggestion some more

misi9170:
@Bartdoekemeijer, all intended functionality is now in. The new example (in examples/layout/turbine_dependencies.py) should give an idea of functionality. @paulf81 does this hit all the functionality you'd like to see?

Bartdoekemeijer:
I actually have been using the `get_all_impacting_turbines` function for a new thing I have been working on internally and it's been great so far! The examples look good too. Thanks for doing this Misha!

paulf81:
Hi Misha this is great and indeeds include all the functionality I hoped for in a really neat arrangement!  I added a few small comments (maybe check I am indeed explaining nomenclature right).  Otherwise this is good to merge from my perspective.

paulf81 (approved):

misi9170:
@paulf81 Thanks, yes, your comments look good to me. Squashing and merging.

PR #71, opened 2023-03-04
Feature/allow pass in axarr

paulf81:
I think it is

Small update to allow for axarr to be passed into plotting functions.  This would be useful for instance if you want to declare the figure outside the function, or plot into a subplot of other axis.  If axarr is not passed in, functionality should be unchanged.

None

Energy ratio plotting functions and calling functions

I reran the energy ratio examples and they all seemed to work the same as before.


misi9170 (approved):
Looks good, thanks!

paulf81:
Thanks Misha!  Merging...

PR #72, opened 2023-03-07
Feature: Energy ratio polar plots and wind speed distributions

Bartdoekemeijer:
This PR adds two useful functionalities to the energy ratio plotting library:

1. The possibility for polar plots. This is based on an article by K. Gunn, "Improvements to the Eddy Viscosity Wind Turbine Wake Model", where the power ratios are plotted over a polar coordinate system.
2. The plotting of the wind speed bins in the energy ratio bin plots. In the current plots, all wind speed bin counts are thrown on a single pile and then plotted. It would be extremely insightful to also see the wind speed distribution within each energy ratio. We do this now. This immediately makes the bottom plot better represent the actual wind conditions.

Combining (1) and (2) turns the bin count plot into a proper wind rose plot. Here's an example from the `flasc_cookiecutter_template` repository with the new functionality:
image

image

None.

Energy ratio classes and energy ratio visualization classes. No fundamental changes in how energy ratios are calculated or used.

N/A.

### DIFF
--- a/examples/energy_ratio/energy_ratio_vs_wd_for_multiple_dfs.py
+++ b/examples/energy_ratio/energy_ratio_vs_wd_for_multiple_dfs.py
@@ -122,4 +122,5 @@ if __name__ == "__main__":
         percentiles=[5., 95.],
         verbose=False)
     fsc.plot_energy_ratios(superimpose=True)
+    fsc.plot_energy_ratios(superimpose=True, polar_plot=True)  # Also show in a polar plot
     plt.show()
--- a/examples/energy_ratio/energy_ratio_vs_wd_for_single_df.py
+++ b/examples/energy_ratio/energy_ratio_vs_wd_for_single_df.py
@@ -95,6 +95,10 @@ if __name__ == '__main__':
     ax[0].set_title("Energy ratios for turbine 001 without UQ")
     plt.tight_layout()
 
+    fig, ax = era.plot_energy_ratio(polar_plot=True)  # Plot in polar format too
+    ax[0].set_title("Energy ratios for turbine 001 without UQ")
+    plt.tight_layout()
+
     # Get energy ratio with uncertainty quantification
     # using N=50 bootstrap samples and 5-95 percent conf. bounds.
     era.get_energy_ratio(
--- a/flasc/energy_ratio/energy_ratio.py
+++ b/flasc/energy_ratio/energy_ratio.py
@@ -495,6 +495,9 @@ class energy_ratio:
         Args:
             hide_uq_labels (bool, optional): If true, do not specifically label
                 the confidence intervals in the plot
+            polar_plot (bool, optional): Plots the energy ratios in a polar
+                coordinate system, aligned with the wind direction coordinate
+                system of FLORIS. Defaults to False.
 
         Returns:
             ax [plt.Axes]: Axis handle for the figure.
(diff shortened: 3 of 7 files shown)


misi9170:
@Bartdoekemeijer nice! What version of matplotlib are you running? I found that I needed to update so that ax.legend() would accept the "ncols" keyword argument, and once updated, I found that ax.grid() no longer accepted the "b" keyword argument (this seems to have been replaced with "visible"). I have gone ahead and changed the "b" to "visible" and committed to your branch. This seems to run fine for me now with the latest matplotlib (3.7.1). (I was on 3.5.3 before, an the "ncols" keyword argument was failing). 

We don't specify an exact version for matplotlib in the setup/requirements, which to me is fine---users should just grab the latest.

misi9170:
@Bartdoekemeijer @paulf81 I see that there is now a merge conflict due to fig no longer being passed out from vis.plot()---Bart, are you fine with also removing the plt.gcf() return from vis.plot()? I'll let you manage the conflict, but I'll approve now anyway.

misi9170 (approved):

paulf81:
This is really cool @Bartdoekemeijer  thanks for pushing this up!  I downloaded the code and ran the two modified examples and those looked nice (following upgrading matplotlib like Misha, maybe we should boost the requirement?)

Sorry about the merge conflict!  I can help with that if you like.  Then also, does the example wake_steering_example.py work for you?  I'm getting an error on this one thinking it could be to do with the new bar plots per wind speed?

Then final comment, when we plot two dfs, do you think we should still have side-by-side bars as an option to show balancing?

Bartdoekemeijer:
Thanks for the feedback @misi9170 @paulf81! I have updated the requirements to use at least my version of `matplotlib>=3.6.3` and removed `plt.gcf()` from the function return statement and now to only return the `axarr` variable. Also, I don't have any issues with the wake steering examples in `flasc`, and there are no wake steering examples in the public `flasc_cookiecutter_template` repo. Which example are you refering to @paulf81?

Bartdoekemeijer:
> Then final comment, when we plot two dfs, do you think we should still have side-by-side bars as an option to show balancing?

That function still exists. The thing now is that it reduces to single bars if the two underlying distributions are the same. I think keeping it makes sense because it can give a good insight into the underlying distributions.

Also, I merged the latest develop branch and things should be good to merge now, if you all agree.

paulf81:
> 

HI @Bartdoekemeijer when I try running: examples/energy_ratio/wake_steering_example.py

I get an error:

ValueError: shape mismatch: objects cannot be broadcast to a single shape.  Mismatch is between arg 0 with shape (29,) and arg 1 with shape (20,).

Bartdoekemeijer:
@paulf81 good catch, OK this should do it now! I do wonder how to deal with the bar plots when using `superimpose=True`, because I think they will be plotted on top of one another currently, but that's a logical result of using `superimpose=True` I guess.

paulf81:
hi @Bartdoekemeijer it runs for me now, thanks!  So looking at this plot:

image

What if we mapped which df each bar belonged to a color, and then split wind speeds on alpha?  Then maybe we could have it both ways?

Bartdoekemeijer:
@paulf81 how about this, using colors to correspond the energy ratios to the bar plots and using hatching to distinguish between wind speed bins? 

image

Bartdoekemeijer:
And for the wake steering example, it becomes:
image

misi9170:
@paulf81 yes, go for it

Bartdoekemeijer:
Hold on. I still have to push up these actual commits!

misi9170:
@Bartdoekemeijer haha oops, thanks

PR #73, opened 2023-03-17
Add data plot to visualization

paulf81:
Yes, maybe we can delete demo code after you approve and before merging

data_plot was a function I'd found useful in the past for having an on hand way to represent noisy SCADA data (typical example might be achieved offset).  It exists in a pretty out-dated version in FLORIS, so I've added this updated version to FLASC, and will submit another pull request to FLORIS to remove the outdate code

visualization.py

### DIFF
--- a/flasc/visualization.py
+++ b/flasc/visualization.py
@@ -650,7 +650,7 @@ def plot_binned_mean_and_ci(
     show_bin_points=True,
     show_confidence=True,
     alpha_scatter=0.1,
-    confidenceLevel = 0.95,
+    confidence_level = 0.95,
 ):
     """
     Plot data to a single axis.  Method
@@ -721,11 +721,12 @@ def plot_binned_mean_and_ci(
     df_agg = df_agg[df_agg["y_count"] > 0]
 
     # Add the confidence interval of the mean to df_agg
-    df_agg["y_confidence"] = st.t.interval(confidenceLevel, 
-                                           df_agg["y_count"]-1,
-                                           loc=df_agg["y_mean"],
-                                            scale=df_agg["y_sem"])[1] - df_agg["y_mean"]
-
+    df_agg["y_ci_lower"], df_agg["y_ci_upper"] = st.t.interval(
+        confidence_level, 
+        df_agg["y_count"]-1,
+        loc=df_agg["y_mean"],
+        scale=df_agg["y_sem"]
+    )
 
     # Plot the mean values
     ax.plot(df_agg.x_bin, df_agg.y_mean, color=color, label=label)
@@ -734,8 +735,8 @@ def plot_binned_mean_and_ci(
     if show_confidence:
         ax.fill_between(
             df_agg.x_bin,
-            df_agg.y_mean - df_agg.y_confidence,
-            df_agg.y_mean + df_agg.y_confidence,
+            df_agg.y_ci_lower,
+            df_agg.y_ci_upper,
             color=color,
             alpha=0.2,
         )
@@ -743,14 +744,14 @@ def plot_binned_mean_and_ci(
         # Plot a dasshed line at confidence interval
         ax.plot(
             df_agg.x_bin,
-            df_agg.y_mean - df_agg.y_confidence,
+            df_agg.y_ci_lower,
             color=color,
             alpha=0.2,
             ls="--",
         )
         ax.plot(
             df_agg.x_bin,
-            df_agg.y_mean + df_agg.y_confidence,
+            df_agg.y_ci_upper,
             color=color,
             alpha=0.2,
             ls="--",
--- a/flasc/visualization.py
+++ b/flasc/visualization.py
@@ -14,6 +14,7 @@
 import numpy as np
 from matplotlib import pyplot as plt
 import pandas as pd
+import scipy.stats as st
 
 
 def plot_with_wrapping(
@@ -638,9 +639,7 @@ def label_line(
         raise ValueError("Need one of near_i, near_x, near_y")
 
 
-
-
-def data_plot(
+def plot_binned_mean_and_ci(
     x,
     y,
     color="b",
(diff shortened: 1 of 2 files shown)


paulf81:
Noting related FLORIS pull: https://github.com/NREL/floris/pull/606

flasc/visualization.py
```diff
@@ -639,12 +640,137 @@ def put_label(i):
...
+        {"y": ["count", "std", "min", "max", "mean"]}
+    )
+    # Flatten column names
+    df_agg.columns = ["_".join(c) for c in df_agg.columns]
+
+    # Reset the index
+    df_agg = df_agg.reset_index()
+
+    # Add the confidence interval of the mean to df_agg
+    df_agg["y_confidence"] = 1.96 * df_agg["y_std"] / np.sqrt(df_agg["y_count"])
```
Bartdoekemeijer, line 719: Where does this equation come from? Can we add a variable called `confidence_interval` to allow the user to specify the confidence interval?

flasc/visualization.py
```diff
@@ -639,12 +640,137 @@ def put_label(i):
...
+            color=color,
+            s=df_agg.y_count / df_agg.y_count.max() * 20,
+            alpha=0.5,
+            marker='s'
+        )
+
+    return ax
     
+# Small little demo of data plot, can delete
+if __name__ == "__main__":
```
Bartdoekemeijer, line 768: This is a nice example. It would be good to apply this plot somewhere in the `flasc_cookiecutter_template` repository on actual data so that users are made aware of its existence, perhaps?

Bartdoekemeijer (commented):
Nice functionality! See my comments. Cheers!

flasc/visualization.py
paulf81: Yes, that's a good idea, I think 1.96 is the value for 95% confidence but the general equation I think we can add a more generic based on this:

http://www.stat.yale.edu/Courses/1997-98/101/confint.htm#:~:text=The%20critical%20value%20for%20a,%3D%20(100.86%2C%20102.78).

or maybe a more canned approach like:

https://stackoverflow.com/questions/15033511/compute-a-confidence-interval-from-sample-data

?

flasc/visualization.py
paulf81: That is a good idea, just gave myself a todo item to do that, maybe also we can use someplace in flasc's examples...

misi9170 (changes requested):

flasc/visualization.py
```diff
@@ -639,12 +640,137 @@ def put_label(i):
 
 
 
+def data_plot(
```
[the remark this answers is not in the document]
paulf81: How about plot_binned_mean_and_ci ?

flasc/visualization.py
paulf81: Agree, thinking that could be a seperate issue in the cookie_cutter repo?  (To add an example?). I made a small one in the wake steering example but I think we have a broader discussion of where examples go so this could ultimately join that

flasc/visualization.py
paulf81: Thanks guys!  I found a way to do this generally online (using st but a slightly different implementation) but I think this one is maybe now wrong, it's using a t-distribution, what do you think?  Maybe I should've grabbed your approach @misi9170

misi9170: @paulf81 I think what you have, using the t-distribution, is good (it's a little more correct than the z-score approach, which assumes the data set is large enough that the means can be treated as normal---the t-distribution doesn't make that assumption, and approaches the z-score method as the number of samples becomes large). If rearranged the code a little in my commits, but have not changed the method.

misi9170 (approved):

flasc/visualization.py
```diff
@@ -637,14 +639,137 @@ def put_label(i):
...
+    # Delete rows with no data
+    df_agg = df_agg[df_agg["y_count"] > 0]
+
+    # Add the confidence interval of the mean to df_agg
+    df_agg["y_ci_lower"], df_agg["y_ci_upper"] = st.t.interval(
+        confidence_level, 
+        df_agg["y_count"]-1,
+        loc=df_agg["y_mean"],
+        scale=df_agg["y_sem"]
+    )
```
misi9170, line 729: @paulf81 I think what you have, using the t-distribution, is good (it's a little more correct than the z-score approach, which assumes the data set is large enough that the means can be treated as normal---the t-distribution doesn't make that assumption, and approaches the z-score method as the number of samples becomes large). If rearranged the code a little in my commits, but have not changed the method.

flasc/visualization.py
```diff
@@ -637,14 +639,137 @@ def put_label(i):
...
+    # Declare ax if not provided
+    if ax is None:
+        _, ax = plt.subplots()
 
+    # Put points ino dataframe
+    df = pd.DataFrame({"x": x, "y": y})
 
+    # If x_edges not provided, use 50 bins over range of x
+    if x_edges is None:
+        x_edges = np.linspace(df["x"].min()*.98, df["x"].max()*1.02, 50)
```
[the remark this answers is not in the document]
misi9170: I do see that the warnings that I was trying to avoid are still being raised in the example, I think because the data isn't really spread along the x axis in a realistic way. For now, I don't think that's really a problem.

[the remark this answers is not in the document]
Bartdoekemeijer: If you are using the example dataset, absolutely the data is mostly in really low wind speeds, indeed unrealistic. This has been an annoyance for me while working with this data, for sure. I wasn't sure if we wanted to fix this or just wait for the revision of the examples to finish.

paulf81:
Hi @Bartdoekemeijer  and @misi9170 what do you think, good to merge this one?  I know we have some open conversations but unless I misread we can put off full solutions to these to other activities?

misi9170:
> Hi @Bartdoekemeijer and @misi9170 what do you think, good to merge this one? I know we have some open conversations but unless I misread we can put off full solutions to these to other activities?

@paulf81, good by me.

paulf81:
Squashing and a'merging

PR #76, opened 2023-03-22
Add conversion from char to rgb

paulf81:
Ready to be merged

Had a small issue after the polar plotting pull request, a case I was running specified color to energy ratio plots via simple characters like 'k' or 'r', adding the .to_rgb command I think is harmless if already rgb, but converts to a tuple if supplied a character, and this avoids an error when this line in energy_ratio_visualization.py is run:

```
bar_color = colors[ii] + alpha_array[wsii] * (1.0 - np.array(colors[ii]))
```
Since you can't subtract a string from a float

### DIFF
--- a/flasc/energy_ratio/energy_ratio_visualization.py
+++ b/flasc/energy_ratio/energy_ratio_visualization.py
@@ -110,6 +110,9 @@ def plot(
         # If nothing specified, use the default tableau colors from matplotlib.pyplot
         colors = [mcolors.to_rgb(clr) for clr in mcolors.TABLEAU_COLORS.keys()]
 
+    # In case colors is specified by a character, convert to RGB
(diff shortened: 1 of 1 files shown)


misi9170 (approved):
Looks fine to me @paulf81! I checked that it worked when specifying colors using `'C0'`, `'C1'`, etc, (which is just an easy way to get at matplotlib's default colors in a desired way) and it worked fine for those, too.

paulf81:
> Looks fine to me @paulf81! I checked that it worked when specifying colors using `'C0'`, `'C1'`, etc, (which is just an easy way to get at matplotlib's default colors in a desired way) and it worked fine for those, too.

Thanks for checking that!

Bartdoekemeijer (approved):
Looks good to me, nice catch! Feel free to merge.

PR #86, opened 2023-05-30
Bug fix: various small corrections

Bartdoekemeijer:
This PR is not yet ready to be merged.

Various bug fixes throughout the code.

I found various issues running FLASC for wind farm SCADA analyses.

Various

### DIFF
--- a/flasc/utilities.py
+++ b/flasc/utilities.py
@@ -13,7 +13,7 @@
 
 import datetime
 
-import numba
+# import numba
 import numpy as np
 # import scipy.interpolate as interp
 
@@ -79,7 +79,7 @@ def interp_with_max_gap(x, xp, fp, max_gap, kind, wrap_around_360=False):
 
 # Credits to 'np8', from https://stackoverflow.com/questions/64045034/interpolate-values-and-replace-with-nans-within-a-long-gap
 # Adapted to include nearest-neighbor interpolation
-@numba.njit()
+# @numba.njit()
 def _interpolate_with_max_gap(
     x, xp, fp, max_gap, assume_sorted=False, kind="linear", extrapolate=True,
 ):
--- a/flasc/dataframe_operations/dataframe_filtering.py
+++ b/flasc/dataframe_operations/dataframe_filtering.py
@@ -133,7 +133,6 @@ def df_mark_turbdata_as_faulty(df, cond, turbine_list, exclude_columns=[]):
         turbine_list = [turbine_list]
 
     for ti in turbine_list:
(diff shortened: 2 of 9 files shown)


paulf81 (approved):
I had some feedback at WESC there was a little trouble setting up FLASC so perhaps related to these issues you identify Bart, maybe we can call this a hot fix, pull to main and do some 0.01 increment in version?

Bartdoekemeijer:
Yes let's do that! There might be a couple more bugs that I haven't found yet though. I'm working on FLASC a lot over the next 2 weeks. Is it OK if we keep this PR open until then? That way I can add on as I find more bugs.

paulf81:
> Yes let's do that! There might be a couple more bugs that I haven't found yet though. I'm working on FLASC a lot over the next 2 weeks. Is it OK if we keep this PR open until then? That way I can add on as I find more bugs.

Sounds good to me!  I will also be tinkering (polars and database stuff), I just caught my polars pull up to yours and will try to keep up

Bartdoekemeijer:
Sorry I had made an absolute mess mixing up some PRs here. Restored it with a force push.

Bartdoekemeijer:
In the most recent commit, I vectorized some code that was previously in a `for` loop. This helped speed up the part where we calculate the histogram/median power curve based on the data. Additionally, I have reduced the calls to this function to save time, and only call it when it's actually needed. The `ws_pow_filtering` code now runs about a factor 5 times faster on my machine.

flasc/utilities.py
```diff
@@ -13,7 +13,7 @@
 
 import datetime
 
-import numba
```
rafmudaf, line 16: Is it necessary to maintain this as a commented code or ok to delete? It seems that numba will not be used so, in my opinion, leaving these lines in but commented means its out of context.

Bartdoekemeijer: `numba` has been a big source of dependency errors and code inconsistency. However, when it did work, it sped up certain time resampling operations a lot. I'm not currently looking at any high-resolution data, but would be worth reconsidering this option when I do.

Bartdoekemeijer:
OK to merge into `develop` if you ask me!

Bartdoekemeijer:
Just committed the PyPI bug fix from @rafmudaf.

paulf81:
@rafmudaf do you approve merging?  If so you can either merge or let me know and I'll do it, thanks!!

rafmudaf:
@paulf81 Yes, from the perspective of the `setup.py` infrastructure, this is good to go.

paulf81:
Merged!

PR #91, opened 2023-06-21
replace append with concat

paulf81:
Elizabeth found a bug in FLASC that if you install now, pandas 2.0 is installed, and it no longer supports the append statement.  This causes an error in the example: flasc/examples_smarteole/05_baseline_energy_ratio_analysis.ipynb

This bugfix resolves the issue by replacing the deprecated append, with concat


misi9170:
> Perhaps there should be a separate PR on upgrading to Pandas 2.0? Also, I did replace a bunch of append statements in #86.

Thanks for pointing that out Bart! Agreed; let's see what @paulf81 would like to do.

paulf81:
That works for me too, I was mainly hoping to "hotfix" this one since without the correction some of the smarteole examples have errors, but can also do all at once?  Do we think we could bang this out quick?

Issue #100, opened 2023-07-07
[Feature]: Plot uncorrected and corrected energy ratios to show impact of northing calibration


paulf81:
### Description

Provide a standard tool to plot the energy ratio using the uncorrected wd and the northing corrected wd to demonstrate where the wd has been corrected to

misi9170:
Working on this this week, will provide a link to a pull request once generated.

misi9170:
Now addressed (I hope) in #105

misi9170:
Closed by #105

PR #105, opened 2023-08-10
Option to show uncorrected energy ratio in northing calibration plots

misi9170:
Ready for review and merge.

Issue #100 was raised that it would be nice to allow users to plot the bias-uncorrected energy ratio along with the corrected one for bias calibration. This pull request addresses that issue, I think.

Checklist:
- [X] Implement source code
- [x] Update relevant examples
- [x] Check tests pass


paulf81 (approved):
Implementation looks good, I downloaded the code and ran the examples and they look good, and that this does produce the expected result, I approve!

Issue #104, opened 2023-07-22
Decide on binning conventions for FLASC


paulf81:
### Description

I was hoping we could use this issue to make a decision on binning conventions in FLASC.  Or maybe we have one already and I just need to stick to it!  But just in case, I had a proposal that can be incorporated within PR #102 where we use the following conventions everywhere:

 - Conventions apply to binning ws, wd, power, and time
 - If you supply a minimum value, a maximum value, and a step size, the binning assumes you gave the left most and right most edge, and the labels will be the center of every bin (even in time)
 - Or should time be an exception?  If it is I prefer left-edge labels, although in FLASC I think we've used right-edge
 - Every function that accepts a min, max and step should alternatively accept the bin_edges directly, but the bin_labels will yet be created as before
 - bins will not be an input because it can be more ambiguous than bin_edges
 - overlapped bins accomplished by reducing the bin size to some fraction of the original request and using a rolling operation instead of a groupby.  Amount of overlap specified using an input like percentage overlap

Interested to hear what you all think!

Bartdoekemeijer:
Thanks for starting this discussion! A lot of things have been implicit in the code and never properly formalized/thought out, so happy that we're doing it now.

> Conventions apply to binning ws, wd, power, and time

Should we add TI as a variable to bin over?

> If you supply a minimum value, a maximum value, and a step size, the binning assumes you gave the left most and right most edge, and the labels will be the center of every bin (even in time)

Agreed

> Or should time be an exception? If it is I prefer left-edge labels, although in FLASC I think we've used right-edge

I don't have a preference here -- as long as we are consistent throughout.

> Every function that accepts a min, max and step should alternatively accept the bin_edges directly, but the bin_labels will yet be created as before. bins will not be an input because it can be more ambiguous than bin_edges.

Agreed


> overlapped bins accomplished by reducing the bin size to some fraction of the original request and using a rolling operation instead of a groupby. Amount of overlap specified using an input like percentage overlap

I like this. So it'll default to `1.0` to use it normally and users can set it to `10` for example to have a bin width of 30 deg for a wd bin step of 3 deg?

paulf81:
Thanks for these helpful responses @Bartdoekemeijer , @ejsimley I thought I'd just quick check if you any comment coming from conventions in openOA we can match?  Otherwise I think we can press on, maybe start by opening an issue capturing these changes we intend to make

paulf81:
Thank you Bart!  This is helpful responses!  A few more from me,

[...]

PR #102, opened 2023-07-12
Feature/port to polars 3

paulf81:
Not ready to be merged

**Feature or improvement description**
This is a final reboot of the port_to_polars efforts, initially begin in Pull Request #81 and updated in Pull Request #97, here following the decision made in discussion #98 that we keep the assumption that pandas dataframe defined in the current methods (with columns pow_000, pow_001, and time) will be kept the same, and conversion to polars can happen within certain functions.

In connection with this effort is a small repo https://github.com/paulf81/flasc_metrics (based on https://github.com/rafmudaf/floris_metrics) which can be used to confirm that even with a pandas->polars->pandas round trip the functions are still sped up.  The new folder timing_tests provides functions to that repo.

A first result is promising, converting the energy ratio with bootstrapping to the polars-based method reduces the average time to compute an N=20 bootstrap energy ratio over the artificial data set from 11.3344 seconds to 1.0762 seconds.

**Related issue, if one exists**
Issue #80 

**Impacted areas of the software**
energy_ratio
(this list may grow)

PR #97, opened 2023-07-03
Add polars based energy ratio

paulf81:
Not ready to be merged

**Feature or improvement description**
Starting the port to polars pull request (#81) over again from the latest develop branch and moving some code to new files where it's not clear whether we do or don't want to port.  Will also open a discussion on this topic.  Basically for code that is not slow, or not run repeatedly, perhaps it's not worth the port?

PR #81, opened 2023-04-24
Feature/port to polars

paulf81:
Not ready to be merged

**Feature or improvement description**
This pull request, currently very much just a function-by-function work in progress to replace pandas code with polars to facilitate faster computations, allow multiprocessing and improve memory usage.

Since this is a going through, also taking the opportunity to make some side-changes to sqlalchemy linked code, and move some os -> pathlib, but these are mainly minor changes.

**Related issue, if one exists**
Issue #80 

**Impacted areas of the software**
Will be a fairly pervasive change.

PR #107, opened 2023-08-17
Feature/port to polars 4

paulf81:
Port to polars (4th try)

This is a final, final, reboot of the port_to_polars efforts, initially begin in Pull Request #81 and updated in Pull Request #97, (and then pull request #102) here following the decision made in discussion #98 that we keep the assumption that pandas dataframe defined in the current methods (with columns pow_000, pow_001, and time) will be kept the same, and conversion to polars can happen within certain functions.  Here I also adjust to discussion Issue #104 

Unlike the previous versions I also split some things apart I think in a way that adds more clarity.  

energy_ratio_utilities.py contains helper functions other modules could need
energy_ratio.py includes the main working functions of energy ratio calculation
energy_ratio_result.py defines a class meant to hold the energy ratio results in a convenient way and provide clear plotting functions.

Unlike before I've included methods for overlapping bins and now you can the tests energy_ratio_test.py that I've defined for polars passes giving identical results.  In general I've confirmed that the energy ratio result values match all previous results except a small but intentional change in how rebalancing is done.  This change only manifests when using noisy data and more than one set of data (for example baseline and wakesteering) and is still generally small.  But for this reason the tests pass with precision of 4 decimal places.

I'm temporary leaving the polars implementations split out during development but hope to move quickly to deleting the old energy_ratio codes once all surrounding codes are edited to use the new functions.  Should hopefully go quick.

Finally you'll also see I'm playing with default visualizations, happy for any and all suggestions here too!

In connection with this effort is a small repo https://github.com/paulf81/flasc_metrics (based on https://github.com/rafmudaf/floris_metrics) which can be used to confirm that even with a pandas->polars->pandas round trip the functions are still sped up.  The new folder timing_tests provides functions to that repo.

A first result is promising, converting the energy ratio with bootstrapping to the polars-based method reduces the average time to compute an N=20 bootstrap energy ratio over the artificial data set from 11.3344 seconds to 1.0762 seconds.

Issue #80 

energy_ratio
(this list may grow)


paulf81:
Updated the test code in https://github.com/paulf81/flasc_metrics and still seeing a 10x reduction in time taken in computation

Bartdoekemeijer:
Great stuff! Let me know when it makes sense for me to review.

paulf81:
With luck you should be able to review tomorrow, thank you @Bartdoekemeijer !

misi9170:
> Great stuff! Let me know when it makes sense for me to review.

@Bartdoekemeijer At this point you can go ahead and review! The artificial data examples are not yet updated (I'll be working on those tomorrow), but the Smarteole examples are ready to roll and we don't expect any further fundamental changes to the energy_ratio/ code.

paulf81:
@Bartdoekemeijer ok, I think we're ready to merge on our end.  If you want to have a look around, all examples updated to new syntax.  We've also added this notebook as quicker tour: https://github.com/paulf81/flasc/blob/feature/port_to_polars_4/examples_artificial_data/energy_ratio/00_demo_energy_ratio_syntax.ipynb

PR #122, opened 2023-09-08
Feature/add options on count

paulf81:
Ready to be merged

This pull request changes the current behavior that assumes, when weighting wind speed bins in the energy ratio across dataframes, that we should use the minimum count (thus up-weighting the wind-speed bins that have the most points in the lesser case), allow an option to weight by sum.  Minimum will stay the default but now an option to pass in 'sum' to a new weight_by variable.

The pull request includes a new test which shows and checks the functionality on a small problem

Finally, a small unrelated bug in energy_ratio_output (remove_all_bins used instead of remove_all_nulls is fixed) and the energy_ratio example for wake steering is cleaned up.

This is in support of functions to compute total uplift

New test added test_alternative_weighting, all tests pass

### DIFF
--- a/flasc/energy_ratio/energy_ratio.py
+++ b/flasc/energy_ratio/energy_ratio.py
@@ -39,6 +39,7 @@ def _compute_energy_ratio_single(df_,
                          ws_min = 0.0,
                          ws_max = 50.0,
                          bin_cols_in = ['wd_bin','ws_bin'],
+                         weight_by = 'min', #min, sum
                          wd_bin_overlap_radius = 0.,
                          uplift_pairs = [],
                          uplift_names = [],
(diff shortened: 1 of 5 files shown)


misi9170 (approved):

flasc/energy_ratio/energy_ratio.py
```diff
@@ -186,6 +193,10 @@ def _compute_energy_ratio_bootstrap(er_in,
         ws_min (float): The minimum wind speed to use.
         ws_max (float): The maximum wind speed to use.
         bin_cols_in (list[str]): A list of column names to use for the wind speed and wind direction bins.
+        weight_by (str): How to weight the energy ratio, options are 'min', or 'sum'.  'min' means
+            the minimum count across the dataframes is used to weight the energy ratio.  'mean' means the mean
+            count across the dataframes is used to weight the energy ratio.  'sum' means the sum of the counts
```
ejsimley, line 198: I think this line about 'mean' means the mean count across... should be removed. It doesn't seem like the 'mean' option is implemented anywhere.

flasc/energy_ratio/energy_ratio_output.py
```diff
@@ -58,6 +59,9 @@ def __init__(self,
             ws_min (float): The minimum wind speed value.
             ws_max (float): The maximum wind speed value.
             bin_cols_in (List[str]): TBD
+            weight_by (str): How to weight the energy ratio, options are 'min', , or 'sum'.  'min' means
```
ejsimley, line 62: there's an extra comma

ejsimley (approved):
Just a couple comments on the doc strings. Otherwise looks good.

Bartdoekemeijer:
Seems very useful!

flasc/energy_ratio/energy_ratio_output.py
paulf81: Fixed

PR #123, opened 2023-09-11
Feature/provide frequency

paulf81:
Could be ready to merge

This pull request enables the bin weighting in the energy ratio calculation to come from an externally provided pandas dataframe of weights, rather than necessarily the count of points in the binned data.  The assumed implementation provides a dataframe with ws/wd columns matched to the labels inferred from ws_min, ws_step, wd_min etc.,

I had originally assumed that if this table was passed in then the ws/wd bins/edges could be inferred from this table.  But in the process of writing the tests realized that an edge cases of single ws or wd might make a challenge here (although solvable) so went back to basically this table just has to match what is otherwise provided by the ws_min etc., parameters.  A potential gotcha is for example the user specifies ws_min =0, ws_step = 1, then the bins will be 0.5, 1.5 etc., maybe we can provide helpful warnings?

Otherwise seems to work as is but wanted to maybe use the pull request to discuss how we best handle some of these finer points.


paulf81:
Maybe an idea @misi9170 you had was a pre-processor to generate the table, we can pass in an FI wind rose or a table, plus the ws_min/max/step and project so that it will line up with what energy_ratio expects?

flasc/energy_ratio/energy_ratio.py
```diff
@@ -297,6 +316,10 @@ def compute_energy_ratio(er_in: EnergyRatioInput,
         weight_by (str): How to weight the energy ratio, options are 'min', , or 'sum'.  'min' means
             the minimum count across the dataframes is used to weight the energy ratio.   'sum' means the sum of the counts
             across the dataframes is used to weight the energy ratio.
+        df_freq (pd.Dataframe): A dataframe which specifies the frequency of the ws/wd bin combinations.  Provides
+            a method to used an explicit or long-term weigthing of various bins.  Dataframe should include
+            columns ws, wd and freq_val.  ws and wd should correspond to the bin centers resulting from
+            the choices of the ws/wd_min / _max / _step.  Defatuls to None
```
ejsimley, line 322: Suggest describing how the weights are handled when the df_freq bins don't match the bins given by ws/wd min, max, step perfectly. So if df_freq has extra bins that aren't included in those given by ws/wd min, max, step, they will be ignored in the energy ratio calculation. And any bins given by ws/wd min, max, step not present in df_freq will be assigned a frequency of zero. At least I think that's how those situations are handled.

flasc/energy_ratio/energy_ratio.py
```diff
@@ -381,6 +405,29 @@ def compute_energy_ratio(er_in: EnergyRatioInput,
     # Convert the numbered arrays to appropriate column names
     test_cols = [f'pow_{i:03d}' for i in test_turbines]
 
+    # If df_freq is provided, confirm is consistent with ws/wd min max and
+    # prepare a polars table of weights
+    if df_freq is not None:
+
+        # Maybe not test, not sure yet
```
ejsimley, line 412: I think it would be good to check the match between the ws/wd min, max, step bins and those in df_freq here. Maybe this could raise an error if none of the bin centers given by ws/wd min, max, step are included in df_freq, and print a warning message if some but not all of the ws/wd min, max, step bins are included in df_freq.

flasc/energy_ratio/energy_ratio.py
```diff
@@ -105,28 +105,43 @@ def _compute_energy_ratio_single(df_,
...
+    
+    # Determine the weighting of the ws/wd bins
+
+    if df_freq_pl is None:
+        # Determine the weights per bin as either the min or sum count
+        df_ = (df_
+            .with_columns(
+                [
+                    # Get the weighting by counts
+                    pl.col('count').min().over(bin_cols_without_df_name).alias('count_weight') if weight_by == 'min' else
```
ejsimley, line 130: could "count_weight" be changed to "weight" throughout the code? Especially if df_freq is used, the the weights aren't really related to counts.

ejsimley (changes requested):
Overall, looks good Paul. Just a few comments.

I didn't follow this statement, though: "A potential gotcha is for example the user specifies ws_min =0, ws_step = 1, then the bins will be 0.5, 1.5 etc., maybe we can provide helpful warnings?" What is the problem with this case?

flasc/energy_ratio/energy_ratio.py
paulf81: This is a good idea, I think why I had paused this earlier and left as a todo is what is the best way to check?  Should I just make sure that at least 1 wd bin and 1 ws bin match?  Or should I confirm that we have a match on at least 1 (wd,ws) combination?  Maybe that is the best approach, convert the unique ws/wd combinations into a list of tuples and do a set difference?  Then check the length, if 0 raise error, < the length of the ws/wd min, max, step raise warning?

flasc/energy_ratio/energy_ratio.py
```diff
@@ -105,32 +105,47 @@ def _compute_energy_ratio_single(df_,
...
+                ]
+            )
         )
+    
+    else:
+        # Use the weights in df_freq_pl directly
+        df_ = (df_.join(df_freq_pl, on=['wd_bin','ws_bin'], how='left')
+              .with_columns(pl.col('weight').fill_null(strategy="zero"))
+        )
+
```
paulf81, line 141: @ejsimley, reviewing your comments made me realize a possible issue.  Say the weights coming from df_freq assign weight to a bin for which one of the cases has 0 points.  In this case at least we need to remove the bin right?  I can make a test for this case, maybe this happens anyway and I'm not realizing it, but maybe we also need to explicitly check it right?

flasc/energy_ratio/energy_ratio.py
ejsimley: I hadn't thought about it, but now that you mention the options, I think it makes sense to check for matches on the (wd,ws) combinations, and the error vs. warning logic you described sounds like a good approach to me.

flasc/energy_ratio/energy_ratio.py
ejsimley: I think you're right, but now I'm wondering if assigning a weight from df_freq to a bin where only one of the cases has data is any different than if we used the "sum" weighting option for the same scenario. Either way, there would be a non-zero weight applied to a bin with data in only one case. I guess that wound't make sense when computing the uplift for two cases and it would be best to exclude the bin entirely. It should be ok if a user is only interested in the individual energy ratios, but we could assume that if someone's not interested in computing the uplift, they'd just create energy ratios with only one case at a time.

So I vote for setting the weight to zero if data are missing for any of the cases in a particular bin, though not sure the best way to do that.

Bartdoekemeijer:
Something I had before was a 2D interpolant function that would take the wind direction and wind speed, and would return the frequency of occurrence. This way, you could calculate the energy ratios using a 20-year-equivalent wind rose even with a smaller amount of data. I thought the interpolant function was particularly helpful because it's agnostic to the binning options chosen. That code snippet is here: https://github.com/NREL/flasc/blob/88330a767d1a91dff49fcef39ea7d1ce67ae526d/flasc/energy_ratio/energy_ratio.py#L45.

Is this more or less what you want to reproduce now with Polars? Sorry for being late to the conversation.

flasc/energy_ratio/energy_ratio.py
paulf81: This is a good catch, I just pushed up a fix that ensures that all ws/wd bins appear in all dataframes, and added a test that this works as expected

flasc/energy_ratio/energy_ratio.py
paulf81: Ok, believe I've now caught this case more generally and added a test to confirm

paulf81:
> Something I had before was a 2D interpolant function that would take the wind direction and wind speed, and would return the frequency of occurrence. This way, you could calculate the energy ratios using a 20-year-equivalent wind rose even with a smaller amount of data. I thought the interpolant function was particularly helpful because it's agnostic to the binning options chosen. That code snippet is here:
> 
> https://github.com/NREL/flasc/blob/88330a767d1a91dff49fcef39ea7d1ce67ae526d/flasc/energy_ratio/energy_ratio.py#L45
> 
> .
> Is this more or less what you want to reproduce now with Polars? Sorry for being late to the conversation.

This sounds really nice Bart!  Right now I'm just assuming you are passing in a dataframe with the same ws/wd bins identified as in the data.  But I like your suggestion, what if for simplicity though we handled this option as a new feature in a separate pull request?  I can see expanding on this capability as we go,

Bartdoekemeijer:
> What if for simplicity though we handled this option as a new feature in a separate pull request? I can see expanding on this capability as we go,

Sounds good!

paulf81:
> > What if for simplicity though we handled this option as a new feature in a separate pull request? I can see expanding on this capability as we go,
> 
> Sounds good!

Ok opened Issue #125 to capture this idea / plan

paulf81:
Ok @ejsimley / @misi9170 @Bartdoekemeijer I think this is now ready to merge from my perspective if 1-2 of you could provide a review.  Thank you!

misi9170:
> Maybe an idea @misi9170 you had was a pre-processor to generate the table, we can pass in an FI wind rose or a table, plus the ws_min/max/step and project so that it will line up with what energy_ratio expects?

Yeah, agreed, just documenting our discussion from today---let's do that, but create a separate PR for it.

paulf81:
> > Maybe an idea @misi9170 you had was a pre-processor to generate the table, we can pass in an FI wind rose or a table, plus the ws_min/max/step and project so that it will line up with what energy_ratio expects?
> 
> Yeah, agreed, just documenting our discussion from today---let's do that, but create a separate PR for it.

Done!  Created Issue #126

misi9170:
@paulf81 Working on creating an extra part of the energy_ratio examples to specify `df_freq`. Hoping to have that done today, and should be able to provide a review at that time.

misi9170:
@paulf81 Have now revamped how df_freq is computed if it is not passed in, passed that through to the Output object, added option to overlay the frequency on the wind direction count plot and shown how to use it in Smarteole examples. 

Overall looking good to me, going to approve.

misi9170 (approved):

paulf81:
thank you @misi9170, these changes look really great and clean things up a lot!  @ejsimley back to you now for review when you are available

tests/energy_ratio_test.py
```diff
@@ -272,12 +350,76 @@ def test_alternative_weighting(self):
...
+
+        df_wake_steering  = pd.DataFrame({'wd': [270, 270., 270.,270.,],
+                           'ws': [7., 7., 8.,8.],
+                           'pow_000': [1., 1., 1., 1.],
+                           'pow_001': [2., 2., 1., 1.],
+        })
+
+        er_in = EnergyRatioInput([df_base, df_wake_steering],['baseline', 'wake_steering'], num_blocks=1)
+
+        # In the final test, specify a bin frequency where 7 m/s is 90% and 8 m/s is 10%
```
ejsimley, line 371: Update the comment to "specify uniform bin frequencies"

flasc/energy_ratio/energy_ratio.py
```diff
@@ -115,30 +119,38 @@ def _compute_energy_ratio_single(df_,
...
+    # Check if all the values in the weight column are null
+    if df_['weight'].is_null().all():
+        raise RuntimeError("None of the ws/wd bins in data appear in df_freq")
+    
+    # Check if any of the values in the weight column are null
+    if df_['weight'].is_null().any():
+        warnings.warn('Some bins in data are not in df_freq and will get 0 weight')
 
+    # Fill the null values with zeros
+    df_= df_.with_columns(pl.col('weight').fill_null(strategy="zero"))
```
ejsimley, line 154: I don't think this affects any of the computed energy ratio values, so it likely isn't necessary, but what do you think about normalizing 'weight' so it sums to 1? The only place where this might have an impact is when overlaying weights on the energy ratio plots.

ejsimley (approved):
This looks good to me! Sorry for the delay in reviewing. Just a few small documentation comments, and some questions/ideas that aren't necessary to address now. 

One thing to fix, though, is that there is an error in the "Comparison assuming no wind direction uncertainty/variability in FLORIS" section of the example 06 notebook.

flasc/energy_ratio/energy_ratio.py
paulf81: Maybe like this?

flasc/energy_ratio/energy_ratio.py
```diff
@@ -115,30 +119,38 @@ def _compute_energy_ratio_single(df_,
...
     df_ = (df_
         .filter(pl.all_horizontal(pl.col(bin_cols_with_df_name).is_not_null())) # Select for all bin cols present
-        .groupby(bin_cols_with_df_name, maintain_order=True)
+        .group_by(bin_cols_with_df_name, maintain_order=True)
         .agg([pl.mean("pow_ref"), pl.mean("pow_test"),pl.count()])
-        )
-    
+
+        # Enforce that each ws/wd bin combination has to appear in all dataframes
+        .filter(pl.count().over(bin_cols_without_df_name) == num_df)
```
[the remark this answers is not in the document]
paulf81: I think this implementation is the most general, say you want to compare a data partioned into 3 stability types, this checks that each type has at least one point in the used bins.  I think it's a fair qualifier.

tests/energy_ratio_test.py
paulf81: good catch

Issue #133, opened 2023-11-14
[BUG] Polars versioning


paulf81:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Having a small issue with polars versioning where version above 0.19.5 are not able to run on my mac for reasons related to:

https://github.com/pola-rs/polars/issues/12454

And starting with 0.19.13, CI is failing for a polars-test fail, but I can't debug locally because of the earlier issue.  Would like to find a longer term solution, but for now setting the requirement to 0.19.5 and I think we can revisit in a few weeks.

### Environment

```markdown
- OS:
- pip environment (can be retrieved with `pip list`):
```

paulf81:
Maybe a way to deal with this is to make a flag to our install, such if you pip install with [mac], it installs the LTS version, and installs the normal build otherwise?  Does that sound reasonable @rafmudaf ?

rafmudaf:
That sounds like a workable approach, assuming there's a corresponding description in the installation documentation, for a developer install. However, this isn't an option for installing from PyPI.

paulf81:
Going to close this issue as FLASC has updated to polars v1 and seems to work

PR #140, opened 2023-12-04
Feature/update reqs pin polars

### LINKED ISSUE #139: Change dependent versioning to semantic versioning using the compatible release specifier
### Description

Improve dependencies to use compatible release specifier such that the dependencies won't go up a major version unless intentionally changed in an update to this repo. This is to avoid breaks due to API changes. Info on compatible release specifier:

https://packaging.python.org/en/latest/specifications/version-specifiers/#id4

### Related URLs

_No response_

paulf81:
Ready to be merged

This pull request implements two small changes to the requirements:

- Changes dependent versioning to use latest compatible version (semantic versioning) [Issue #139]
- Reinforces that we're back-pinning polars until we agree on general [Issue #133 ]

Except for polars, requirements specified via ~=1.1, which is interpreted as >=1.1, ==1.*, which essentially means latest minor version which matches major version, and is >= minor version specified.  Can see:

https://packaging.python.org/en/latest/specifications/version-specifiers/#id4

Closes #139 
Issue #133 

setup.py

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,7 +12,7 @@ AUTHOR = "NREL National Wind Technology Center"
 
 # What packages are required for this module to be executed?
 REQUIRED = [
-    'bokeh~=3.1.1',
+    'bokeh~=3.1',
     'floris~=3.4',
     'feather-format~=0.0',
     'ipympl~=0.9',
--- a/setup.py
+++ b/setup.py
@@ -12,21 +12,20 @@ AUTHOR = "NREL National Wind Technology Center"
 
 # What packages are required for this module to be executed?
 REQUIRED = [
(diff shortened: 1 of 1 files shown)


misi9170 (approved):
I've cloned this, installed it, and run the artificial data examples without any issues so I think you're good to go here. 

Note that this is PR is currently pointing to merge into NREL:main, which I think is what you intended because it's a bug fix, but just checking

paulf81:
> Note that this is PR is currently pointing to merge into NREL:main, which I think is what you intended because it's a bug fix, but just checking

Thanks @misi9170 !  That's right, plan was to merge into main and then back merge this change to develop, I'll do it now...

Issue #138, opened 2023-11-27
[Documentation]: FLASC could use additional examples of calculating total uplift


paulf81:
### Description

This is currently I think only in one of the smarteole examples, some dedicated examples would help clarify syntax

christiannvaughn:
Should we close this issue? @paulf81 @misi9170

misi9170:
> Should we close this issue? @paulf81 @misi9170

Yes, please do!

PR #155, opened 2023-12-18
Feature/add uplift example

paulf81:
This pull request adds an additional example demonstrating the calculation of total uplift.  I think it's mostly ready but there are a few finer points I'd like to work through before merging:

- [ ] What are the warnings about?  Can they be squashed
- [ ] Should we lock in a random seed to guarantee one result so the documentation can be more specific?

Closes Issue #138

### DIFF
--- a/examples_artificial_data/03_energy_ratio/05_wake_steering_example.py
+++ b/examples_artificial_data/03_energy_ratio/05_wake_steering_example.py
@@ -42,7 +42,6 @@ if __name__ == "__main__":
     )
     num_points_per_combination = 5  # 5 # How many "seconds" per combination
 
-    # I know this is dumb but will come back, can't quite work out the numpy version
     ws_array = []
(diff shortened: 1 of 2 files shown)


misi9170:
Nice! On your points above:
- I'm strongly in favor of adding `np.random.seed(0)` right after `if __name__ == "__main__":` for reproducibility
- Did you mean to leave `plt.show()` off? There are a lot of plots made in this script that are abandoned
- I'm still looking into the warnings, but they all originate from the `plot_binned_mean_and_ci` call, so if you don't want to plot anything anyway that line can be commented out and the warnings don't appear. On the other hand, it's probably best to track those down and see what causes them?

misi9170 (approved):

paulf81:
Thanks @misi9170 !  I know you're making some last checks but feel free to merge when done, thank you!

PR #164, opened 2024-02-02
Add coveralls to CI

paulf81:
Small pull request to add coveralls (https://coveralls.io/) to FLASC CI.  Coveralls provides line-by-line checks on coverage and produces a report on test coverage.

PR #166, opened 2024-02-03
Try codecov as alternaitve

paulf81:
Try alternative to #164 with codecov to see which we like better


paulf81:
Finally think I prefer coveralls so closing this one

Issue #175, opened 2024-03-04
`find_sensor_stuck_faults` only returns indices of faults for the last column


DanielKerrigan:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

The [find_sensor_stuck_faults](https://github.com/NREL/flasc/blob/71516963d299a22a0f6201080d398b2c52cf0626/flasc/turbine_analysis/find_sensor_faults.py#L20) function has a parameter for a list of columns, but it only returns the stuck indices for the last column in the list.

### Expected Behavior

My expectation is that the function would return the indices where the values in any of the columns are stuck and that the order of the columns in the list would not matter.

### Steps To Reproduce

```python
import pandas as pd
from flasc.turbine_analysis.find_sensor_faults import find_sensor_stuck_faults

df = pd.DataFrame({"a": [0, 0, 0, 1, 2, 3], "b": [4, 5, 6, 7, 7, 7]})

results_a = find_sensor_stuck_faults(df, columns=["a"], ti=0, plot_figures=False)
print("columns=['a']:", results_a)

results_b = find_sensor_stuck_faults(df, columns=["b"], ti=0, plot_figures=False)
print("columns=['b']:", results_b)

results_ab = find_sensor_stuck_faults(df, columns=["a", "b"], ti=0, plot_figures=False)
print("columns=['a', 'b']:", results_ab)

results_ba = find_sensor_stuck_faults(df, columns=["b", "a"], ti=0, plot_figures=False)
print("columns=['b', 'a']:", results_ba)

# Output:
# columns=['a']: [0 1 2]
# columns=['b']: [3 4 5]
# columns=['a', 'b']: [3 4 5]
# columns=['b', 'a']: [0 1 2]
```

### Environment

```markdown
- OS: MacOS 14.2.1
- pip environment:
    - flasc 1.5
```

misi9170:
Hi @DanielKerrigan, thank you for raising this issue. I've run your example script (which is very useful, thank you for providing it) and confirmed the outputs your are seeing. I've also tracked down the issue in `find_sensor_stuck_faults` and will work on a bugfix in the coming days.

PR #177, opened 2024-03-05
[BUGFIX] Fix stuck sensor detection only applying to last column

misi9170:
Addresses issue raised in #175.

The stuck sensor indices were only being returned for the last column checked, as opposed to a union of the indices over all columns. This PR addresses that issue. 

### To do
- [X] Address fundamental issue and ensure that indices are returned for any stuck sensor (not just final column)
- [x] Check that examples still run as expected
- [x] Add test for this (can base on @DanielKerrigan 's example script)
- [x] to discuss: do we want to return a separate list of indices for each column provided, or simply a union of all indices that contain a stuck sensor fault in at least one column (the current expected behavior)? _I decided to add an option to do this, defaulting to_ `False`_, which when_ `True` _will instead return a dictionary containing a numpy array of stuck sensor indices for each column passed._ 

@DanielKerrigan 's script below now produces the expected results
```python
import pandas as pd
from flasc.turbine_analysis.find_sensor_faults import find_sensor_stuck_faults

df = pd.DataFrame({"a": [0, 0, 0, 1, 2, 3], "b": [4, 5, 6, 7, 7, 7]})

results_a = find_sensor_stuck_faults(df, columns=["a"], ti=0, plot_figures=False)
print("columns=['a']:", results_a)

results_b = find_sensor_stuck_faults(df, columns=["b"], ti=0, plot_figures=False)
print("columns=['b']:", results_b)

results_ab = find_sensor_stuck_faults(df, columns=["a", "b"], ti=0, plot_figures=False)
print("columns=['a', 'b']:", results_ab)

results_ba = find_sensor_stuck_faults(df, columns=["b", "a"], ti=0, plot_figures=False)
print("columns=['b', 'a']:", results_ba)

# Output:
# columns=['a']: [0 1 2]
# columns=['b']: [3 4 5]
# columns=['a', 'b']: [0 1 2 3 4 5]
# columns=['b', 'a']: [0 1 2 3 4 5]
```


misi9170:
Two of the examples (that I'm aware of) use `find_stuck_sensor_faults()`, called within ws_pow_filtering.py's `filter_by_sensor_stuck_faults()`:
- examples_artificial_data/01_raw_data_processing/00_filter_ws_power_curves.ipynb
- examples_smarteole/03_filter_ws_power_curves.ipynb
In these, `find_stuck_sensor_faults()` is being called _separately for each turbine_ in a `for` loop; however, the turbine's `wd` and `ws` columns are checked simultaneously, in that order, so the bug meant that only the stuck `ws` indices were being returned. After the bugfix, there is a slight change (increase) to the data removed by the stuck sensor filter for examples_artificial_data/01_raw_data_processing/00_filter_ws_power_curves.ipynb; and there is no (or minimal) change to the data removed by the stuck sensor filter in examples_smarteole/03_filter_ws_power_curves.ipynb (presumably, there were few/no instances of stuck `wd` readings).
Out of caution, I have rerun any subsequent notebooks after these examples so that they are consistent with any changes to the data made.

paulf81 (approved):
These changes look good @misi9170, and thank you for adding the tests

misi9170:
Thanks @paulf81 . I realized I hadn't pushed the changes to the examples (not to the running code, just to the notebook output), so I'll do that now and then merge.

Issue #171, opened 2024-02-08
[BUG] Fix plot_layout in energy_ratio_het_mapper


paulf81:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Somewhere in the v2 conversion, maybe based on updates to package dependencies, example 05, which runs the het_mapper started throwing errors in plotting related to lin interpolant.  Logging this issue to remember to fix it before release.  Current output:

```
  File "/Users/pfleming/Projects/FLASC/flasc/examples_artificial_data/03_energy_ratio/05_estimate_heterogeneity_from_energy_ratios.py", line 88, in <module>
    hm.plot_layout(plot_background_flow=True, ylim=[0.90, 1.10], pdf_save_path=pdf_save_path)
  File "/Users/pfleming/Projects/FLASC/flasc/flasc/analysis/energy_ratio_heterogeneity_mapper.py", line 323, in plot_layout
    lin_interpolant = LinearNDInterpolator(
  File "interpnd.pyx", line 290, in scipy.interpolate.interpnd.LinearNDInterpolator.__init__
  File "_qhull.pyx", line 1827, in scipy.spatial._qhull.Delaunay.__init__
  File "_qhull.pyx", line 353, in scipy.spatial._qhull._Qhull.__init__
scipy.spatial._qhull.QhullError: QH6154 Qhull precision error: Initial simplex is flat (facet 1 is coplanar with the interior point)

While executing:  | qhull d Qz Qc Q12 Qt Qbb
Options selected for Qhull 2019.1.r 2019/06/21:
  run-id 2042442214  delaunay  Qz-infinity-point  Qcoplanar-keep
  Q12-allow-wide  Qtriangulate  Qbbound-last  _pre-merge  _zero-centrum
  Qinterior-keep  Pgood  _max-width 6.2e+03  Error-roundoff 6.6e-12
  _one-merge 4.6e-11  Visible-distance 1.3e-11  U-max-coplanar 1.3e-11
  Width-outside 2.6e-11  _wide-facet 7.9e-11  _maxoutside 5.3e-11

The input to qhull appears to be less than 3 dimensional, or a
computation has overflowed.

Qhull could not construct a clearly convex simplex from points:
- p69(v4): 4e+02 7.1e+02     0
- p30(v3): 2.9e+03 -7.1e+02 1.4e+03
- p0(v2): 4.8e+03 -1.8e+03 4.3e+03
[...]

PR #189, opened 2024-04-10
bug fix in plot_layout

paulf81:
This quick bugfix is meant to address Issue #171 by catching a certain qhull error when using the lin interpolator and falling back to 100% NN interpolator (which is already used to plug nans).  @Bartdoekemeijer  just wanted to flag for you but this will probably be a done deal when you see it...


misi9170 (approved):
Works for me

PR #199, opened 2024-07-25
Add new `FlascDataFrame` object to enhance user interface with data

misi9170:
`FlascDataFrame` inherits from `pandas.DataFrame`, but will provide some extra utilities for mapping between different naming conventions and dataframe formats ("long" data, "wide" data, etc).

At this stage, just a proof of concept while we build out needed functionality, tests, etc.


misi9170:
This is now clean enough that I'm going to merge into the flasc-dataframe branch, which will still be a work in progress while we build out more features on the `FlascDataFrame`

paulf81 (approved):

PR #204, opened 2024-08-13
Update docs to describe PyPI installation.

misi9170:
The [installation page of the docs](https://nrel.github.io/flasc/installation.html) incorrectly states that FLASC must be cloned and installed locally. This PR fixes that and describes installation from PyPI.


paulf81 (approved):

paulf81:
Nice catch!

Issue #195, opened 2024-06-05
Incorporate wind-up features into FLASC


paulf81:
### Description

The wind-up repo (https://github.com/resgroup/wind-up) by @aclerc provides a set of tools to calculate yield uplift of wind turbines.  This issue is meant to track an effort to develop a module within flasc that calls wind-up functions from within the flasc framework.  Per the approach we've been taken more and more lately, a draft PR will be opened to track the progress of the work and open up potential collaboration on implementation.

paulf81:
This was completed within #212

PR #210, opened 2024-09-04
Add wind up example using `FlascDataFrame`

aclerc:
Add a simple example of a [wind-up](https://github.com/resgroup/wind-up) uplift and uncertainty calculation
The filtered and northed flasc dataframe is converted to wind-up format using a new `FlascDataFrame` method
It is probably best to merge PR #199 before merging this PR because I branched from `misi9170:feature/flasc-df`

#195 

I have added `test_convert_to_windup_format()` which passes for me locally


misi9170:
@aclerc I've merged back the flasc-dataframe target branch and resolved conflicts. Tests pass, and I checked that the examples_smarteole/09_wake_steering_uplift_analysis.ipynb notebook still runs and appears to produce the expected outputs. 

This remerging may need to happen one more time if we get #211 , which adds some core components to the `FlascDataFrame`, goes in before this PR, but I again should be able to handle that.

aclerc:
Hi @misi9170 regarding `OptionError: No such keys(s): 'future.no_silent_downcasting'` I think that is related to the pandas option set on this line https://github.com/resgroup/wind-up/blob/eee1c07be49e17d344feaa4aa0f4af8e14ff4648/wind_up/plots/scada_funcs_plots.py#L220. I think upgrading your pandas version will fix that for you. If this continues to be troublesome feel free to raise an issue on the wind-up repo.

misi9170:
Thanks @aclerc, that fixed it. I was using pandas 2.0.2 when I was having the problem, and upgrading to the current 2.2.2 solved the problem (I also had to upgrade pyarrow for compatibility with pandas 2.2.2). pandas 2.0.2 actually satisfies the FLASC pandas requirement (`pandas~=2.0`), so I'll probably bump the requirement up to `pandas~=2.2`. 

Confirming now that the notebook produces the expected 0.6%, -0.2%, 1.4% result.

paulf81 (approved):

PR #211, opened 2024-09-05
Add dataframe conversions to flascdataframe

paulf81:
# Add dataframe conversions to flascdataframe
This pull request adds in the conversions from flasc format to user format, and relavent testing.  

Converts from flasc format to supported wide and long user formats and adds additional tests.

Remaining question: Do we want any dedicated documentation of these conversions?

### DIFF
--- a/flasc/flasc_dataframe.py
+++ b/flasc/flasc_dataframe.py
@@ -7,76 +7,76 @@ from pandas import DataFrame
 class FlascDataFrame(DataFrame):
     """Subclass of pandas.DataFrame for working with FLASC data.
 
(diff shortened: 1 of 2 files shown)


paulf81:
@misi9170 I think I've implemented this all now according to the scheme we laid out and would now like your review.  Note that I did switch back to requiring the long_column_names as being the means by which the user format is determined.

paulf81:
@misi9170 , a medium-sized update this afternoon, I realized I wasn't testing the cases where the channel_map doesn't specifically denote how to change each channel, added those tests and fixed the bugs they revealed

PR #229, opened 2024-10-28
[BUGFIX] Fix pypi action to not use setup.py

paulf81:
Fix the pypthon-publish action to work with pyproject.toml


misi9170:
I think we should point this at develop, and then do the normal develop -> main release afterwards. I could do that immediately, and then release v2.1.1, or we can leave it and bundle it with other patches before a 2.1.1 release. As it stands, v2.1 is now successfully available on PyPI (https://pypi.org/project/flasc/) and I've confirmed that I can install and access it via the usual `pip install flasc`.

PR #236, opened 2024-11-06
Implement expected analysis

paulf81:
# Implement expected analysis

This branch will implement the methods of uplift analysis described in [AWC validation methodology](https://publications.tno.nl/publication/34637216/LWOb3s/TNO-2020-R11300.pdf).  It also makes some changes to names of existing functions to better clarify the operation of the seperate methods for analyzing data and quantifying uplift.

Changes to be included:

- [x] Renaming `EnergyRatioInput` to `AnalysisInput` to make clear can be used for both
- [x] Providing an alias to `EnergyRatioInput` with deprecation warning to avoid breaking changes
- [x] Renaming `ComputeTotalUplift` to `TotalUpliftPowerRatio` to clarify the method of analysis
- [x] Add an alias with deprecation warning
- [x] Updating all references in code
- [x] Adding new module `expected_power_analyisis.py` which will implement the methods in [AWC validation methodology](https://publications.tno.nl/publication/34637216/LWOb3s/TNO-2020-R11300.pdf)
- [x] Add new module `expected_power_analysis_output.py` for catching the output of functions in `expected_power_analysis` and provided post-processing operations
- [x] Adding tests for any new function
- [x] Add an example
- [x] Adding documentation


flasc/analysis/expected_power_analysis.py
```diff
@@ -0,0 +1,767 @@
...
+        remove_all_nulls_wd_ws (bool): Remove all null cases for wind direction and wind speed.
+             Defaults to False.
+        remove_any_null_turbine_bins (bool): When computing farm power, remove any bins where
+            and of the test turbines is null.  Defaults to False.
+        remove_any_null_turbine_std_bins (bool): Remove bins from consideration where any of
+            the covariance or num_points columns are null.  Defaults to False.
+        variance_only (bool): Only use the variance of turbine powers and not turbine-turbine
+            covariances. Defaults to False.
+        fill_cov_with_var (bool): Fill all missing covariance terms with the product of the
+            square root of the variances. Defaults to False.
```
ejsimley, line 586: Reminder to raise an error if both `variance_only` and `fill_cov_with_var` are true. When both are true it's not really  clear which option gets used.

flasc/analysis/expected_power_analysis.py
```diff
@@ -0,0 +1,767 @@
...
+        use_standard_error (bool): Use standard error for the uplift calculation. Defaults to True.
+        N (int): The number of bootstrap samples. Defaults to 1.
+        percentiles (List[float]): The percentiles to calculate for the bootstrap samples.
+            Defaults to [2.5, 97.5].
+        remove_all_nulls_wd_ws (bool): Remove all null cases for wind direction and wind speed.
+             Defaults to False.
+        remove_any_null_turbine_bins (bool): When computing farm power, remove any bins where
+            and of the test turbines is null.  Defaults to False.
+        remove_any_null_turbine_std_bins (bool): Remove bins from consideration where any of
+            the covariance or num_points columns are null.  Defaults to False.
```
ejsimley, line 582: I think we could get rid of this `remove_any_null_turbine_std_bins` argument and always keep the bins where covariance is missing. This was important before we added the following two arguments, but now that we have options for filling in or zeroing missing covariances, we'll still be able to compute uncertainty in this case.

paulf81: Removed!

flasc/analysis/expected_power_analysis.py
paulf81: Done!

flasc/analysis/expected_power_analysis.py
```diff
@@ -0,0 +1,757 @@
...
+    ws_step: float = 1.0,
+    ws_min: float = 0.0,
+    ws_max: float = 50.0,
+    bin_cols_in: List[str] = ["wd_bin", "ws_bin"],
+    weight_by: str = "min",  # min or sum
+    df_freq: pd.DataFrame = None,
+    use_standard_error: bool = True,
+    N: int = 1,
+    percentiles: List[float] = [2.5, 97.5],
+    remove_all_nulls_wd_ws: bool = False,
```
paulf81, line 536: @ejsimley should I go ahead and clear remove_all_nulls_wd_ws as it is also too much of an edge case?

ejsimley: Yes, I say remove it to simplify the function, since it probably is an edge case.

paulf81: Done!

ejsimley:
@paulf81, I was going to add a test to make sure the uplift variance is correct using `_total_uplift_expected_power_with_standard_error` but then saw a couple potential issues with the way uncertainty gets calculated. 

First, if setting `variance_only = True` or `fill_cov_with_var = True` so we can still get an uncertainty estimate even when we have missing covariance terms, what happens when some variance terms are Null (because there is only one sample in a bin for the turbine)? Then it seems we still wouldn't be able to get a non-Null uplift variance. Should we remove bins where there is only one sample (or remove individual turbines in the bin from the rest of the analysis if there is only 1 sample)?

The other issue is when `remove_any_null_turbine_bins = False`, which is the default case. If a turbine is missing in a particular bin we can still compute the expected farm power in that bin, we just ignore the missing turbines. But when computing the farm power variance for that bin, I think the variance terms for all of the turbine pairs still get used. So the farm power variance will be Null for all of these cases, even though we intended to just not include the missing turbines for those bins. I think this could be addressed relatively easily when computing farm variance, but wanted to see what you think.  

Also, let me know if I'm wrong about any of the above. I didn't get a chance to thoroughly check this behavior yet.

flasc/analysis/expected_power_analysis.py
```diff
@@ -0,0 +1,738 @@
...
+    # with pl.Config(tbl_cols=-1):
+    #     print(df_cov)
+
+    # Bin and group the dataframe
+    df_bin = _bin_and_group_dataframe_expected_power(
+        df_=df_,
+        test_cols=test_cols,
+        bin_cols_without_df_name=bin_cols_without_df_name,
+    )
+
```
ejsimley, line 348: I was thinking this could be a good place to call a new utility function that synchronizes nulls between df_cov and df_bin (something like "_synchronize_mean_power_cov_nulls"). Specifically, for each row, if there are any turbines in df_cov with undefined variances or covariances (because count < 2), then the mean power for those turbines would get set to Null as well. This way, we would always be able to return a standard error for the uplift by excluding turbines in a given bin with undefined covariance from both the expected uplift and uncertainty calculations. I like this because we wouldn't end up returning a NaN uncertainty value when calling this function. In most cases, though, we would probably just want to set `variance_only = True` or `fill_cov_with_var = True` to maximize the number of turbines that can be used to compute the uplift in each bin.

flasc/analysis/expected_power_analysis.py
```diff
@@ -0,0 +1,738 @@
...
+    # This is equation 4.16
+    df_bin = df_bin.with_columns(
+        pow_farm_var=pl.sum_horizontal(
+            [pl.col(cov) / pl.col(n) for cov, n in zip(cov_cols, n_cols)]
+        )
+    )
+
+    # If any of the cov_cols are null, set pow_farm_var to null
+    df_bin = df_bin.with_columns(
+        pl.when(pl.all_horizontal([pl.col(c).is_not_null() for c in cov_cols]))
```
ejsimley, line 389: Connected to the comment above, I think we could get rid of this requirement that all cov_cols are defined. When computing the "pow_farm_var" column a few lines above, if my understanding is correct, the summation over cov_cols will just ignore any Null values. So, we'd be calculating the variance of the farm power considering only the turbines that are valid, in the same way that "pow_farm" will only sum the power of the turbines that are not Null. Therefore, the farm power variance will correspond to the same set of turbiens used to compute farm power.

flasc/analysis/expected_power_analysis.py
```diff
@@ -0,0 +1,738 @@
...
+        sync_cols=[f"{col}_mean" for col in test_cols],
+        uplift_pairs=uplift_pairs,
+        bin_cols_without_df_name=bin_cols_without_df_name,
+    )
+
+    # Remove rows where any of the test columns are null (if remove_any_null_turbine_bins is True)
+    if remove_any_null_turbine_bins:
+        df_bin = df_bin.filter(
+            pl.all_horizontal([pl.col(f"{c}_mean").is_not_null() for c in test_cols])
+        )
```
ejsimley, line 361: Connected to the comment above, at this point, regardless of the value of "remove_any_null_turbine_bins," we might need to remove rows where all test_cols are Null. Although rare, I think it is possible that if count < 2 for all turbines in a bin, then after synchronizing Nulls between df_cov and df_bin, we could be left with rows that are all Null that should get filtered out from the analysis.

paulf81:
Putting this PR on hold temporarily to start a new branch without covariance calculations.  When we pick this one back up, the issue is that the latest null sync code discussed above, has a theoretical issue in which if a turbine is missing from a bin, it's joint covariance with each turbine will lead all turbines to be set to NULL for that bin.  Need to resolve that issue before moving forward.

flasc/analysis/expected_power_analysis.py
paulf81: This holds now I think so commenting out

flasc/analysis/expected_power_analysis.py
paulf81: Ok I think this is now in the line below

paulf81:
Update: rather than putting this on hold we reconfigured the code so that instead of the options:

variance_only
fill_cov_with_var

There is only the option to do one of these:

set_cov_to_zero_or_var (can be 'zero' or 'var')

And covariances are all overwritten, either to zero, or to the variance products.

Note the variance option keeps the bin counts, while the zero overwrites them (as irrelavalent) and practically speaking this allows more data to be used, while set_cov_to_zero_or_var='var' will likely cause a lot of data to be thrown out for bins where any test turbine is missing, for reasons described above

flasc/analysis/expected_power_analysis_utilities.py
```diff
@@ -0,0 +1,484 @@
...
+                pl.when(pl.col(n_col) < 2).then(None).otherwise(pl.col(n_col)).alias(n_col)
+            )
+
+            # Wherever n_col is null, set cov_col to null
+            df_cov = df_cov.with_columns(
+                pl.when(pl.col(n_col).is_null())
+                .then(None)
+                .otherwise(pl.col(cov_col))
+                .alias(cov_col)
+            )
```
ejsimley, line 363: If n_col is Null, then won't cov_col already be null?

flasc/analysis/expected_power_analysis_utilities.py
```diff
@@ -0,0 +1,484 @@
...
+            pl.when(pl.col(n_col) < 2)  # | pl.col(n_col).is_null())
+            .then(None)
+            .otherwise(pl.col(cov_col))
+            .alias(cov_col)
+        )
+
+        # Wherever n_col is null, set cov_col to null
+        df_cov = df_cov.with_columns(
+            pl.when(pl.col(n_col).is_null()).then(None).otherwise(pl.col(cov_col)).alias(cov_col)
+        )
```
ejsimley, line 269: It seems like some of the null setting in this function is duplicating what is already done in _fill_cov_with_var. We should check to see what is necessary.

flasc/analysis/expected_power_analysis_utilities.py
```diff
@@ -0,0 +1,484 @@
...
+                pl.when(pl.col("null_map"))
+                .then(pl.min_horizontal(n_1_col, n_2_col))
+                .otherwise(pl.col(n_col))
+                .alias(n_col)
+            )
+
+            # For any rows where n_col is < 2 set n_col to null
+            df_cov = df_cov.with_columns(
+                pl.when(pl.col(n_col) < 2).then(None).otherwise(pl.col(n_col)).alias(n_col)
+            )
```
ejsimley, line 355: If any rows with n_col < 2 for covariance terms are getting set to Null, should we also do that for the diagonal variance terms? I couldn't find if/where n_col for the variance terms gets set to Null.

flasc/analysis/expected_power_analysis.py
```diff
@@ -0,0 +1,789 @@
...
+        df_sub = df_sub.pivot(
+            on="df_name",
+            index=bin_cols_without_df_name,  # + ["weight"],
+        )
+
+        # Assign the weight column to be the mean of weight_[uplift_pair[0]]
+        # and weight_[uplift_pair[1]]
+        df_sub = df_sub.with_columns(
+            weight=(pl.col(f"weight_{uplift_pair[0]}") + pl.col(f"weight_{uplift_pair[1]}")) / 2
+        )
```
ejsimley, line 473: Are there situations where the weights could be different for the two uplift pair data frames?

flasc/analysis/expected_power_analysis.py
```diff
@@ -0,0 +1,789 @@
...
+    ws_step: float = 1.0,
+    ws_min: float = 0.0,
+    ws_max: float = 50.0,
+    bin_cols_in: List[str] = ["wd_bin", "ws_bin"],
+    weight_by: str = "min",  # min or sum
+    df_freq_pl: pl.DataFrame = None,
+    percentiles: List[float] = [2.5, 97.5],
+    remove_any_null_turbine_bins: bool = False,
+    set_cov_to_zero_or_var: str = "zero",
+    use_cov_when_available: bool = False,
```
ejsimley, line 257: Since we're only allowing three options for filling in covariance terms, I say we use one argument instead of the two above. It could be cov_terms, with options "zero", "var", and "cov". In the docstring we can describe how the "cov" option will fill in undefined terms with the product of variances.

paulf81: Nice suggestion!  I made teh change

flasc/analysis/expected_power_analysis.py
paulf81: No, not really anyway.  I had originally for this reason put weight as index value (see comment above).  This was the bug yesterday, some how numerically these identical values can can classified as different leading to wrong splits in groups.  I think having a float, especially a computed one, as a grouping index is probably just a bad idea.  This way is a little clunky but I think much safer

flasc/analysis/expected_power_analysis_utilities.py
paulf81: Possibly, I was thinking if things get copied over could be good to double check in that it doesn't hurt at least

flasc/analysis/expected_power_analysis_utilities.py
paulf81: That will happen in _null_and_sync_covariance (currently line 259)

flasc/analysis/expected_power_analysis_utilities.py
paulf81: I wasn't positive, in the tests I made I can make scenarios where n_col is Null but not cov_col (for that reason if you comment out the above block test_fill_cov_with_var_fill_all will fail), but maybe that is not a realistic condition to test for

flasc/analysis/expected_power_analysis_utilities.py
ejsimley: _null_and_sync_covariance will set the variance term to Null, but should the count value also get set to Null as in the above code for the covariance counts?

ejsimley: As we discussed, we can get rid of this check here and instead set any count columns with values equal to 0 (or < 2, either way works) to Null in _compute_covariance.

flasc/analysis/expected_power_analysis_utilities.py
ejsimley: Instead of setting cov to Null when n_col <2 or n_col is Null, which should already be the case, we can add a test to make sure if number of overlapping points is < 2, the computed covariance will be Null for that turbine pair.

flasc/analysis/expected_power_analysis_utilities.py
```diff
@@ -0,0 +1,484 @@
...
+    df_ = (
+        df_.filter(pl.all_horizontal(pl.col(bin_cols_with_df_name).is_not_null()))
+        # Drop rows where all test_cols are null
+        .filter(pl.any_horizontal([pl.col(c).is_not_null() for c in test_cols]))
+        # Select for all bin cols present
+        .group_by(bin_cols_with_df_name, maintain_order=True)
+        # # Compute the mean and standard deviation of each of test_cols
+        .agg(
+            [pl.mean(c).alias(f"{c}_mean") for c in test_cols]
+            + [pl.var(c).alias(f"{c}_var") for c in test_cols]
```
ejsimley, line 78: We can remove variance here since we're only using variance form the separate covariance matrix now.

flasc/analysis/expected_power_analysis.py
ejsimley: OK, interesting, good catch!

flasc/analysis/expected_power_analysis.py
```diff
@@ -0,0 +1,779 @@
...
+    # if variance_only:
+    #     df_cov = _set_cov_to_zero(df_cov, test_cols=test_cols)
+
+    # Apply Null values to covariance and sync across uplift pairs
+    df_cov = _null_and_sync_covariance(
+        df_cov=df_cov,
+        test_cols=test_cols,
+        uplift_pairs=uplift_pairs,
+        bin_cols_without_df_name=bin_cols_without_df_name,
+    )
```
ejsimley, line 351: As discussed, let's remove the _null_and_sync_covariance function. Most of the operations in the function aren't needed since the output of _compute_covariance will already have Nulls in the right places (and we can add a new test to make sure). The synchronization of Null covariance terms between uplift pair dataframes can instead be accomplished by adding all combinations of cov_t1_t2 to the sync_cols list when calling _synchronize_nulls later.

flasc/analysis/expected_power_analysis_utilities.py
```diff
@@ -0,0 +1,484 @@
...
+    )
+
+    # Join the number of points to the covariance matrix
+    df_cov = df_cov.join(df_n, on=bin_cols_with_df_name, how="left")
+
+    # Enforce that each ws/wd bin combination has to appear in all dataframes
+    bin_cols_without_df_name = [c for c in bin_cols_with_df_name if c != "df_name"]
+    num_df = df_["df_name"].n_unique()
+    df_cov = df_cov.filter(pl.len().over(bin_cols_without_df_name) == num_df)
+
```
ejsimley, line 226: Reminder to set count values of 0 to Null in this function.

flasc/analysis/expected_power_analysis_utilities.py
```diff
@@ -0,0 +1,484 @@
...
+    return df_cov
+
+
+def _null_and_sync_covariance(
```
ejsimley, line 230: Reminder to remove this function and instead synchronize Null covariances together with the mean power values.

flasc/analysis/expected_power_analysis_utilities.py
paulf81: Ok done!

flasc/analysis/expected_power_analysis_utilities.py
paulf81: Realizing it's a bit back and forth since in the function above, _get_num_points_pair, I fill nulls with zeros, so just going the other way and setting zeros to null there

flasc/analysis/expected_power_analysis_utilities.py
paulf81: Removed

flasc/analysis/expected_power_analysis.py
paulf81: Agreed removed

flasc/analysis/expected_power_analysis_utilities.py
paulf81: Ok removed

flasc/analysis/expected_power_analysis_utilities.py
ejsimley: Agreed, keeping counts as the actual number of overlapping samples should work fine

ejsimley (approved):

Issue #235, opened 2024-11-05
[Feature]: Method to set wind direction by upstream turbines


ejsimley:
### Description

The `dataframe_manipulations` module includes the function `set_ws_by_upstream_turbines` but there is no equivalent for wind direction. However, field observations have shown that wind directions measured at waked turbines can contain biases relative to unwaked turbines. This is illustrated in the paper [Schepers, J. G., *et al.*, "Analysis of wake measurements from the ECN Wind Turbine Test Site Wieringermeer, EWTW," *Wind Energy*, 2012.](https://onlinelibrary.wiley.com/doi/full/10.1002/we.488), where the authors show that the wind direction measured at a partially waked turbine is biased away from the side where the turbine is waked. This phenomenon could potentially cause the wind direction based on all turbines to be biased. 

To solve this, a method `set_wd_by_upstream_turbines` (and `set_wd_by_upstream_turbines_in_radius`) could be created. Since a wind direction signal is needed to determine the set of upstream turbines, this function could first use the mean wind direction over all turbines to estimate the set of unwaked turbines, then use the unwaked turbines to compute the final wind direction signal.

ejsimley:
Addressed by PR #237

PR #237, opened 2024-11-08
Adding functions to set wind direction by upstream turbines

ejsimley:
This pull request addresses issue #235 by adding functions `set_wd_by_upstream_turbines` and `set_wd_by_upstream_turbines_in_radius` to the `dataframe_manipulations` module. The reason for these functions is that wind directions measured by waked turbines can contain biases due to wake effects. Therefore, it could be advantageous to only use unwaked turbines to estimate the wind direction. These functions first use the average wind direction over all turbines to identify the unwaked turbines. Then the unwaked turbines (or unwaked turbines within the specified radius for `set_wd_by_upstream_turbines_in_radius `) are used to compute the final wind direction signal.

This PR closes issue #235 

The `dataframe_manipulations` module

Tests added in "dataframe_manipulations_test.py"

### DIFF
--- a/flasc/data_processing/dataframe_manipulations.py
+++ b/flasc/data_processing/dataframe_manipulations.py
@@ -370,6 +370,59 @@ def set_wd_by_all_turbines(
     return _set_col_by_turbines("wd", "wd", df, "all", True)
 
 
+def set_wd_by_upstream_turbines(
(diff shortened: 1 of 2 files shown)


paulf81 (approved):
This looks great @ejsimley , nice and clean!

PR #241, opened 2024-12-18
Change how number of turbines is counted

paulf81:
# Change how number of turbines is counted
The current method for counting the number of turbines uses a loop to see when the next turbine doesn't appear.  This created an issue for a current project, where simply counting the number of columns that fit the pattern "pow_###" should provide the equivalent result and avoid the issue.  This PR makes the change and adds a missing back test of the older `get_num_turbines` function.

### DIFF
--- a/flasc/flasc_dataframe.py
+++ b/flasc/flasc_dataframe.py
@@ -102,10 +102,8 @@ class FlascDataFrame(DataFrame):
         """Return the number of turbines in the dataset."""
         self.check_flasc_format()
 
-        nt = 0
-        while ("pow_%03d" % nt) in self.columns:
-            nt += 1
-        return nt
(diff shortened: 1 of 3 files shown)


ejsimley (approved):
This works fine. However, I realized the code still assumes turbines are numbered sequentially when looping through turbines (e.g., `for ti in range(self.n_turbines)`). So maybe for now just stay with the sequential numbering requirement and we can update later if we decide it makes sense.

paulf81:
Ok, do you have a preference?  Merge this in as harmless or save for a more comprehensive change?

ejsimley:
> Ok, do you have a preference? Merge this in as harmless or save for a more comprehensive change?

Good question. I say merge now, but I'll add an issue so we can keep track of this topic.

paulf81:
Merged!

PR #244, opened 2025-02-04
Test of expected power uplift and variance using cov option

ejsimley:
This adds an additional test of `expected_power_analysis.total_uplift_expected_power` to "expected_power_analysis_test.py" that checks the estimated uplift and standard variance of the uplift when using the "cov" option for computing the covariance terms. 

This is a somewhat trivial case because after dropping bins where the variance can't be calculated, only one wind condition bin remains. However, it provides a simple way to confirm that the farm uplift and variance match what is expected based on the TNO report equations. 

tests/expected_power_analysis_test.py

### DIFF
--- a/tests/expected_power_analysis_test.py
+++ b/tests/expected_power_analysis_test.py
@@ -4,6 +4,7 @@ import numpy as np
 import pandas as pd
 import polars as pl
 from polars.testing import assert_frame_equal
+from pytest import approx
 
 from flasc.analysis.analysis_input import AnalysisInput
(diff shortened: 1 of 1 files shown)


paulf81 (approved):
Looks great and a good test to have, thanks Eric!

paulf81:
I think it's to me to merge it then right @ejsimley @misi9170 ?

misi9170:
> I think it's to me to merge it then right @ejsimley @misi9170 ?

yep!

PR #247, opened 2025-02-17
Update numpy req to 2.0

paulf81:
Now the FLORIS develop is on numpy 2 (https://github.com/NREL/floris/pull/1051), it's important for FLASC develop to track the same change so they can be used together in the same environment.

This PR implements changes:
1) Ups numpy req to 2.0
2) Points to FLORIS 4.3 to match the numpy 2 requirement and not go back to older FLORIS with numpy specified >
3) Fixed np.Inf syntax in floris_tools.py

[](https://github.com/NREL/floris/pull/1051)

pyproject.toml
floris_tools.pyt


paulf81:
@misi9170  I think this one is ready, one thing is maybe when release, we should make sure to up FLORIS dependency to FLORIS minor version with numpy 2

misi9170:
@paulf81 The changes to the running code look fine, but I don't love that the floris dependency is now to git rather than PyPI... my preference would be to hold off merging this until FLORIS has been released with its numpy 2.0 requirement?

paulf81:
> @paulf81 The changes to the running code look fine, but I don't love that the floris dependency is now to github rather than PyPI... my preference would be to hold off merging this until FLORIS has been released with its numpy 2.0 requirement?

Ok, yeah, maybe we pull this through develop and main in one fell swoop on release, makes sense

Issue #252, opened 2025-08-01
[BUG] DeprecationWarning


maapasan:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

Deprecation warnings appear in several functions, due to a change of argument names in some libraries.
May give errors in the future

- [ ] [flasc/data_processing/energy_ratio_wd_bias_estimation.py:L118](https://github.com/NREL/flasc/blob/cab01a4d76d970435aceb1e5af3c75efd9cccbf5/flasc/data_processing/energy_ratio_wd_bias_estimation.py#L118) 
DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)
    self.logger.info("  Constructing energy table for wd_bias of %.2f deg." % wd_bias)

- [ ] [flasc/data_processing/energy_ratio_wd_bias_estimation.py:237](https://github.com/NREL/flasc/blob/cab01a4d76d970435aceb1e5af3c75efd9cccbf5/flasc/data_processing/energy_ratio_wd_bias_estimation.py#L235) 
DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)
    + " WD bias: %.3f deg." % wd_bias

- [ ] [flasc/utilities/energy_ratio_utilities.py:580](https://github.com/NREL/flasc/blob/cab01a4d76d970435aceb1e5af3c75efd9cccbf5/flasc/utilities/energy_ratio_utilities.py#L580): 
DeprecationWarning: `pl.count()` is deprecated. Please use `pl.len()` instead.
  (Deprecated in version 0.20.5)
    .agg([pl.mean("pow_ref"), pl.mean("pow_test"), pl.count()])

- [ ] [flasc/utilities/energy_ratio_utilities.py:582](https://github.com/NREL/flasc/blob/cab01a4d76d970435aceb1e5af3c75efd9cccbf5/flasc/utilities/energy_ratio_utilities.py#L582): 
DeprecationWarning: `pl.count()` is deprecated. Please use `pl.len()` instead.
  (Deprecated in version 0.20.5)
    .filter(pl.count().over(bin_cols_without_df_name) == num_df)

[...]

PR #253, opened 2025-08-01
Update numpy, polars, and floris usage to avoid deprecation warnings

misi9170:
As raised in #252 , there are various deprecations warnings raised based on our current requirements (in particular, from numpy and polars). This PR addresses those by fixing.

There seems to be some nuance about when/how deprecation warnings are raised. Simply running the examples as notebooks did not raise the warnings (for me). However, the following workflow works:
- Export the notebook to a python script
- Run the python script with the [`-Wd` flag](https://docs.python.org/3/library/warnings.html#updating-code-for-new-versions-of-dependencies), which explicitly raises deprecation warnings.

To do:
- [x] Address deprecation warnings raised in #252 
- [x] Address new FLORIS warnings raised, where possible
- [ ] Look through examples for other deprecation warnings using the workflow described above
- [ ] Address any other deprecation warnings that arise

### DIFF
--- a/flasc/data_processing/energy_ratio_wd_bias_estimation.py
+++ b/flasc/data_processing/energy_ratio_wd_bias_estimation.py
@@ -89,7 +89,7 @@ class bias_estimation(LoggingManager):
 
     def _load_a_input_for_wd_bias(
         self,
-        wd_bias,
+        wd_bias: float,
     ):
         """Load AnalysisInput objects with bias.
 
@@ -158,7 +158,7 @@ class bias_estimation(LoggingManager):
 
     def _get_energy_ratios_allbins(
         self,
-        wd_bias,
+        wd_bias: float,
         time_mask=None,
         ws_mask=(6.0, 10.0),
         wd_mask=None,
@@ -233,7 +233,7 @@ class bias_estimation(LoggingManager):
 
         for ii, ti in enumerate(self.test_turbines):
             self.logger.info(
-                "    Determining energy ratios for test turbine = %03d." % (ti)
+                "    Determining energy ratios for test turbine = %03d." % ti
                 + " WD bias: %.3f deg." % wd_bias
             )
 
@@ -420,9 +420,14 @@ class bias_estimation(LoggingManager):
         """
         self.logger.info("Estimating the wind direction bias")
 
-        def cost_fun(wd_bias):
+        def cost_fun(x: np.ndarray):
+            """Cost function to minimize.
+
+            Args:
+                x (np.ndarray): Wind direction bias to evaluate. 1D array with 1 element.
+            """
             self._get_energy_ratios_allbins(
-                wd_bias=wd_bias,
+                wd_bias=x[0], # pass as float
                 time_mask=time_mask,
                 ws_mask=ws_mask,
                 wd_mask=wd_mask,
@@ -467,7 +472,7 @@ class bias_estimation(LoggingManager):
             # workers=opt_workers,
         )
 
-        wd_bias = x_opt
+        wd_bias = x_opt[0]
         self.opt_wd_bias = wd_bias
         self.opt_cost = J_opt
         self.opt_wd_grid = x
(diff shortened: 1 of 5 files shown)


misi9170:
Other warnings I'm finding to address (live list):
- [ ] `FutureWarning: The 'axis' keyword in FlascDataFrame.rolling is deprecated and will be removed in a future version. Call the method without the axis keyword instead.`

misi9170:
List from running examples:
- [ ] examples_artificial_data/01.../00.. produces `flasc/data_processing/filtering.py:70: FutureWarning: Setting an item of incompatible dtype is deprecated and will raise an error in a future version of pandas. Value 'nan' has dtype incompatible with bool, please explicitly cast to a compatible dtype first.`

misi9170:
It seems that pandas may be working on a solution for this, a "nullable boolean datatype". The [documentation page](https://pandas.pydata.org/docs/user_guide/boolean.html) indicates that this is experimental, and it doesn't currently avoid the error raised; if I replace the offending line
```python
df.loc[cond, cols] = None 
```
with 
```python
df.loc[cond, cols] = pd.NA
```
the same warning
```
FutureWarning: Setting an item of incompatible dtype is deprecated and will raise an error in a future version of pandas. Value 'nan' has dtype incompatible with bool, please explicitly cast to a compatible dtype first.
```
is raised.

misi9170:
Ok, I've decided to utilize the `exclude_columns` keyword argument in `df_mark_turbdata_as_faulty()` to exclude any boolean type columns from being overwritten. There are only a couple of places where `df_mark_turbdata_as_faulty()` are called, and this seems an appropriate fix for now.

flasc/data_processing/energy_ratio_wd_bias_estimation.py
```diff
@@ -158,7 +158,7 @@ def _load_a_input_for_wd_bias(
 
     def _get_energy_ratios_allbins(
         self,
-        wd_bias,
+        wd_bias: float,
```
paulf81, line 161: why type hint just this one?

flasc/data_processing/time_operations.py
```diff
@@ -56,21 +56,21 @@ def df_movingaverage(
     # Carry out the mean calculations
     df_regular = (
         df[cols_regular]  # Select only non-angular columns
-        .rolling(window_width, center=center, axis=0, min_periods=min_periods)
```
paulf81, line 59: was axis formerly implicit and now has to be explicit?

tests/energy_ratio_test.py
```diff
@@ -225,6 +225,7 @@ def test_flascdataframe_input(self):
 
     def test_energy_ratio_input(self):
         # Energy ratio input is deprecated but should still work
+        # Remove this test in next major release since EnergyRatioInput is deprecated.
```
paulf81, line 228: Maybe just add TODO: flag so it turns up on a global search of TODO

paulf81 (approved):
Looks good @misi9170 , I left a few small comments here and there but all in all looks great.  Thanks for going through all this!

flasc/data_processing/time_operations.py
misi9170: `axis` used to default to 0, but you were allowed to apply the rolling average along the column axis if you liked. This behavior has been deprecated; in future, `rolling` will not accept an axis and will _only_ apply along the index axis. To apply it across columns, the user will have to first transpose the dataframe and then transpose it back.

https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.rolling.html

flasc/data_processing/energy_ratio_wd_bias_estimation.py
misi9170: in `cost_fun`, the argument is a one-element numpy array; however, in `_get_energy_ratios_all_bins` there is some string formatting
```python
" WD bias: %.3f deg." % wd_bias
```
that raises a deprecation warning when `wd_bias` is a numpy array. This update indicates that `wd_bias` should be a float, not a one-element array. For consistency though, I could add type hints to the other arguments here.

paulf81:
> > This sounds reasonable, just to test my understanding, the idea is that all of the work on removing a turbine is when pow_xxx, ws_xxx, wd_xxx are set to nan, having any boolean columns un-flagged won't impact any operations. That works for me but wanted to make sure I understood
> 
> Yeah, I think my _hope_ here is that there aren't places in the code where we use `df.dropna(how="all")`; so that leaving the boolean channels alone won't affect current behavior. I don't see anything like that searching through the code. There are certainly places where we use `dropna` with `"any"` or a subset of the columns (the `float` type ones), so those should not be affected.

I'm on board

